PR #4, opened 2019-02-12
Fixed for supporting flow

ifsnow:
It is the same as the [PR](https://github.com/facebook/react-native/pull/22212) that I submitted in react-native repository.

I hope this helps.

### DIFF
--- a/lib/AsyncStorage.js
+++ b/lib/AsyncStorage.js
@@ -5,8 +5,7 @@
  * LICENSE file in the root directory of this source tree.
  *
  * @format
- * @noflow
- * @flow-weak
+ * @flow
  * @jsdoc
  */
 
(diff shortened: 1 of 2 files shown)


cpojer:
Wow, nice. Thank you! I'll leave it to @Krizzu to review and merge this PR. :)

lib/AsyncStorage.js
```diff
@@ -21,6 +20,21 @@ const RCTAsyncStorage =
   NativeModules.RNC_AsyncSQLiteDBStorage ||
   NativeModules.AsyncLocalStorage;
 
+type ReadOnlyArrayString = $ReadOnlyArray<string>;
```
Krizzu: What about exporting this to different file, say `types/flow.js` and then include it in `.flowconfig`?

ifsnow: @Krizzu 
There is some troubles when using `.flowconfig`. For example, hyperlink(Go to definition) in vscode does not work properly. 

I think the approach below looks ok.

- Added `AsyncStorageType.js` to define types and separated it from `AsyncStorage.js`.
- Added `AsyncStorageType.js.flow` to provide more clean autocomplete.

Let me know what you think about this.

Krizzu: On the second thought, having types defined in the same file is better than have 2 different files with types. Thought extracting flow types into another file would be better supported by VScode.

If you would revert 1 commit back, I'd be more than happy to merge it 👍 
Sorry for confusion!

ifsnow: @Krizzu Sure. I had a similar idea. I modified it as you said.

Krizzu: Great, thank You.
In that case, where we have Types in the same file, do we still need `.js.flow` file? Any advantages to having it?
Also, can you rebase with `master`, please?

ifsnow: @Krizzu With `AsyncStorage.js.flow`, we can use more organized auto-completion in VSCode, but I realized there was a problem with the function to jump directly to `AsyncStorage.js` to check the code.

It seemed to have more disadvantages than merits, so I decided to delete `AsyncStorage.js.flow`. Sorry for confusion!

I finished what you said.

PR #15, opened 2019-02-17
Detox tests

Krizzu:
Add e2e test and run them on CI. 

Todo:

Add test cases:
  - [x] `mergeItem`
  - [x] `removeItem`
  - [x] `clearItem` 

Run detox on CI:
   - [x] iOS
   - [x] Android


example/e2e/asyncstorage.spec.js
```diff
@@ -0,0 +1,142 @@
...
+
+describe('Async Storage', () => {
+  let restartButton;
+  let closeKeyboard;
+  let test_getSetClear;
+  let test_mergeItem;
+
+  beforeAll(async () => {
+    await device.reloadReactNative();
+    restartButton = await element(by.id('restart_button'));
```
TheSavior: Consider adding a test that would fail if restartButton didn’t actually restart. I believe all of the tests in this file would continue passing if restartButton was a no-op.

Issue #13, opened 2019-02-15
Allow control over AsyncStorage location on iOS


cpojer:
* * *
  This issue was originally created by @sam1463 as facebook/react-native#19692.
  * * *

  
  - [x] Review the documentation: https://facebook.github.io/react-native
  - [x] Search for existing issues: https://github.com/facebook/react-native/issues
  - [x] Use the latest React Native version: https://github.com/facebook/react-native/releases
  - [x] Run `react-native info` in your terminal and paste its contents under "Environment"
  - [x] Let us know how to reproduce the issue. Include a code sample, share a project, or
    share an app that reproduces the issue using https://snack.expo.io/

## Environment
Environment:
  OS: macOS High Sierra 10.13.5
  Node: 9.4.0
  Yarn: Not Found
  npm: 5.6.0
  Watchman: 4.9.0
  Xcode: Xcode 9.4.1 Build version 9F2000
  Android Studio: 3.0 AI-171.4443003

Packages: (wanted => installed)
  react: 16.2.0 => 16.2.0
  react-native: 0.52.2 => 0.52.2

## Description

Currently, AsyncStorage stores data in the Documents directory on iOS. However, as discussed in [this issue](https://github.com/facebook/react-native/issues/2477), [this abandoned PR](https://github.com/facebook/react-native/pull/4561), and [this abandoned PR](https://github.com/facebook/react-native/pull/10697), there should be an option to store elsewhere, as this location is not appropriate for non-user-created content, as it is visible for apps that enable file sharing with iTunes or the Files app on iOS.

## Steps to Reproduce
This is the case on AsyncStorage on all versions of react-native to my knowledge.

## Expected Behavior
Some sort of control over the location in which AsyncStorage stores data, so that it will not be exported to iTunes, viewable via the Files app, etc. on iOS

## Actual Behavior
AsyncStorage saves to the Documents directory.

Krizzu:
This will be included in v2 of Async Storage. Please refer to #113 for more details

PR #18, opened 2019-02-23
Added the ability to set the storage location on iOS

adamivancza:
https://github.com/react-native-community/react-native-async-storage/issues/13

Moved my changes from this PR: https://github.com/facebook/react-native/pull/20904/files
I didn't find any tests in the repo atm so I wasn't able to add those. I can see a [PR](https://github.com/react-native-community/react-native-async-storage/pull/15) with detox tests though - if that's merged I can add some tests.


Krizzu:
Hey @adamivancza, thank you for your input!

I'm going to hold off merging this PR till I finish #15 - I'd love to have some test on new functionality, hope you don't mind waiting :)

adamivancza:
sure thing @Krizzu - could you ping me here once you've finished with #15?

Krizzu:
@adamivancza Hey,
Detox is merged, so would be nice if you could add some tests for storage directory changes.

Also, we don't have generated docs from comments yet, so could you please add this method to `docs`, mark it as iOS only and describe its functionality?

Thanks

adamivancza:
thx @Krizzu - will do those over the weekend.

ios/RNCAsyncStorage.m
```diff
@@ -16,9 +16,25 @@
 #import "RCTLog.h"
 #import "RCTUtils.h"
 
+typedef NS_ENUM(NSInteger, RCTStorageLocation) {
+  Documents,
+  ApplicationSupport
```
zhongwuzw: @Krizzu Is this what we want? only support `Documents` and `ApplicationSupport` directories?

Krizzu: It's been discussed in #13, where `ApplicationSupport` is more safe  to store app-specific data.

Do you have any suggestion if there should be more/less options to choose from?

zhongwuzw: @Krizzu Oh, I scan those things, so now I think we may need to handle two things:
1. Provide options for user to save data. (I suppose support three options, `Documents`,`Library`,`tmp`, https://developer.apple.com/library/archive/documentation/FileManagement/Conceptual/FileSystemProgrammingGuide/FileSystemOverview/FileSystemOverview.html#//apple_ref/doc/uid/TP40010672-CH2-SW1, `tmp` is also meaningful IMO, provide temp storage).
2. Not backed up by iTunes. (For the `Application Support` directory, it would backed up by iTunes and iCloud, so if we use that, we need use `NSURLIsExcludedFromBackupKey` for files we stored includes manifest file.

ios/RNCAsyncStorage.m
```diff
@@ -455,4 +489,10 @@ - (NSDictionary *)_writeEntry:(NSArray<NSString *> *)entry changedManifest:(BOOL
   }
 }
 
+RCT_EXPORT_METHOD(setStorageLocation:(RCTStorageLocation)location)
+{
+  storageLocation = location;
```
zhongwuzw: We can have equal check in here.

ios/RNCAsyncStorage.m
```diff
@@ -16,9 +16,25 @@
 #import "RCTLog.h"
 #import "RCTUtils.h"
 
+typedef NS_ENUM(NSInteger, RCTStorageLocation) {
+  Documents,
```
zhongwuzw: Can we conform Apple naming style? 
image

ios/RNCAsyncStorage.m
adamivancza: 1) not sure about tmp - do you have any use cases where it might be helpful? I think that would just cause confusion as tmp dir can be deleted deleted by OS at any time.
2) maybe we can add a flag to control that? it could be useful to have this file backed up.

adamivancza:
Sorry about this but I still haven't got time to finish up the tests - will try to find some time to do them.

adamivancza:
hey @Krizzu - finally finished with the PR! :)

lib/AsyncStorage.js
```diff
@@ -359,6 +359,24 @@ const AsyncStorage = {
       });
     });
   },
+
+  /**
+   * Call this set the storage location.
+   * @param storageLocation The storage location. Can be 'documents' or 'applicationSupport'.
+   *
+   * **Note**: changing this only sets the location for future actions and does not migrate
```
Krizzu: This bit is bit worrying. We had similar issue #55, where storage location was changed and devs were losing data. Maybe we could come up with migration plan for data?

lib/AsyncStorage.js
```diff
@@ -359,6 +359,24 @@ const AsyncStorage = {
...
+   * the store from the previous location.
+   *
+   * @platform ios
+   */
+  setStorageLocationIOS: function(storageLocation: string) {
+    RCTAsyncStorage.setStorageLocation(storageLocation);
+  },
+
+  StorageLocationIOS: {
+    documents: 'documents',
```
Krizzu: I think we could [export constants to be used in JS](https://facebook.github.io/react-native/docs/native-modules-ios#exporting-constants), what do you think?

lib/AsyncStorage.js
adamivancza: The default locations is still the same, imo devs can only loose data once updated to a new version which contains this change AND change the storage location specifically without copying their data.

lib/AsyncStorage.js
adamivancza: oh yeah that's a great shout! will do that

lib/AsyncStorage.js
Krizzu: I see, thanks. We should provide a note in docs that this can happen, so they know upfront.

Krizzu:
Hey @adamivancza, thank you for your work so far 👏 

We merged another PR which contains some native iOS code, I hope this won't give you much trouble.

reilem:
@Krizzu There are some conflicting changes with my fix, mainly with `RCTGetStorageDirectory` and `RCTGetManifestFilePath`. This branch makes some changes to those functions (that I don't entirely agree with). In current master I have made `RCTCreateManifestFilePath` and `RCTCreateStorageDirectoryPath` specifically in case you need a location that isn't the default. I would use these `Create` functions in conjunction with the already existing `RCTGetManifestFilePath` and `RCTGetStorageDirectory`. Otherwise the `dispatch_once` functionality might cause unpredictable behaviour.

Krizzu:
Hey @adamivancza,

Thank you so much for work you put into this PR, but I'm sorry to say that I'm going to close it. We're moving into Async Storage v2, where we do a lot of breaking changes.

There's a plan to still support the current implementation of AS as another Backend Storage in v2 - maybe you'd like to take part in it? We plan to create it from the ground up, so I believe it'd best time to add additional functionality (not preset in AS before)


thanks.

PR #20, opened 2019-02-25
Correct imports

Krizzu:
Currently, iOS is failing on imports, so this fixes the issue.


Krizzu:
Merging this one, as it urgent.

PR #22, opened 2019-02-26
Feature/use async storage

kadikraman:
Fixes https://github.com/react-native-community/react-native-async-storage/issues/16

Adding a `useAsyncStorage` hook to be used with React >=16.7.


Krizzu:
Hi @kadikraman, thanks for your work, looks great!

I'm going to hold off a bit before merging this one, simply because of RN 0.59 (supporting hooks) is not released yet. Also, few notes to consider:

- Can you move this custom hook (and its type) to separate file (say `lib/hooks.js`) please? I'd like to not push everything into one file. 

- Since Hooks are pretty new feature, I think having a separate docs explaining how to use it, with examples would be super useful for new developers


I'm going to assign issue and this PR to you. Sounds good?

kadikraman:
Hi @Krizzu - added a specific example and separated `hooks.js` into its own file. Had to also move `AsyncStorage` to a separate file so I could import it into `hooks.js` without causing a circular dependency, so the entry point is now `index.js`.

docs/API.md
```diff
@@ -493,4 +494,89 @@ static flushGetRequests(): void
...
+
+
+<!-- ------------------------ HOOKS ------------------------ -->
+
+
+## `useAsyncStorage`
+
+Uses the new [hooks api](https://reactjs.org/docs/hooks-intro.html) to give you convenience functions to get, set, merge and delete a value from a location.
+
+The `useAsyncStorage` returns an object that exposes all all the methods that allow you to interact with the stored value.
```
Krizzu: `all all` typo here

docs/API.md
```diff
@@ -493,4 +494,89 @@ static flushGetRequests(): void
...
+    </View>
+  );
+}
+```
+
+In this example:
+
+1. On mount, we read the value at `@storage_key` and save it to the state under `value`
+2. When pressing on "update value", a new string gets generated, saved to async storage, and to the component state
+3. Try refreshing the app with `cmd + R` - you'll see that the last value is still being read from async storage
```
Krizzu: Maybe "Try to reload your app", to keep it platform-agnostic :)

docs/API.md
```diff
@@ -493,4 +494,89 @@ static flushGetRequests(): void
...
\ No newline at end of file
+`undefined`
+
+
+<!-- ------------------------ HOOKS ------------------------ -->
+
+
+## `useAsyncStorage`
+
+Uses the new [hooks api](https://reactjs.org/docs/hooks-intro.html) to give you convenience functions to get, set, merge and delete a value from a location.
```
Krizzu: What do you think about changing `from a location` to `from Async Storage, for given key` or something similar? Just to point that this custom hook is Async Storage key-specific.

kadikraman:
@Krizzu thanks for the feedback - updated.

Krizzu:
Thanks a lot!

vkurchatkin:
Could anyone elaborate, what's the point of this "hook"? It's not even a hook, really, since it doesn't use built-in hooks under the hood.

Krizzu:
To be honest, having a second look at this, I get your point. I definitely see it's clearly missing a 'hook' functionality.

I'll add this to ["clean up" list](https://github.com/react-native-community/react-native-async-storage/issues/21), to make this one actually use hooks API.

Thanks @vkurchatkin

kadikraman:
Yeah, my bad, it doesn't use the built in hooks under the hood. It's just a partially applied function that has a hook-like api.

Maybe a better idea would be to use `useState` to store the current value with some metadata for whether or not it has loaded 🤔

Krizzu:
I've opened an issue for further discussion
#32

PR #26, opened 2019-02-28
rename filenames in code to match case-sensitively real filenames

sylvainbaronnet:


Krizzu:
Looks great 👍 
Can you also commit changed Example app? I believe there're places where it still has old `Asyncstorage` name, so I guess that's worth to update too.

Thanks!

sylvainbaronnet:
I renamed all `Asyncstorage` to `AsyncStorage` including `RNCAsyncStorage.xcodeproj` folder

Issue #33, opened 2019-03-07
Allow brownfield apps to handle requests


tido64:
## Motivation

We have brownfield apps with existing storage solutions that we want our React Native components to take advantage of. For example, one of our apps use [PINCache](https://github.com/pinterest/PINCache) to control lifetime of the stored data and the size of the disk cache.

## Description

Before I start implementing on this, I'd like to run you by my solution and get some feedback on whether I'm on the correct path.

In `RNCAsyncStorage`, we add a delegate `@property`, e.g.:

```objc
@property (nonatomic, weak, nullable) id<RNCAsyncStorageDelegate> delegate;
```

For every `RCT_EXPORT_METHOD`, we call the delegate if it's set, otherwise we do the normal operations.

Consumers can then set the delegate like so:

```objc
RNCAsyncStorage *asyncStorage = [bridge moduleForClass:[RNCAsyncStorage class]];
asyncStorage.delegate = self;
```

Let me know what you think.

Thanks.

Krizzu:
Hey @tido64 	
So, if I understood correctly, you want to expand AS capabilities, by attaching delegates to external storage solutions?

I like the idea. I was thinking about a way to make Async Storage extendable - by applying adapters for external storages.

Are you planning on including this feature in Android too?

tido64:
Yes, that's correct. I haven't looked at the Android implementation yet but we do need it, so I will look at it next when I have time.

tido64:
Merged. Thanks, @Krizzu 🎉

Issue #2, opened 2019-02-11
[General] Update Readme


Krizzu:
Update README to cover:

- Available API (Installation steps, linking)
- Example usages
- Contribution list

Pushplaybang:
@Krizzu  - Readme could also use a "motivation" or "why".

I'm guessing that asyncStorage is not going to be part of core in the future?  couldn't see this in the RN docs, would be great to include a note about this.

Krizzu:
Hey @Pushplaybang,

It's all part of [lean core effort for React Native.](https://github.com/facebook/react-native/issues/23313) I guess it'd be fine to add this info to readme.

PR #35, opened 2019-03-11
feat: Allow brownfield iOS apps to handle storage.

tido64:
Summary:
---------
Adds a delegate property to `RNCAsyncStorage` to allow custom handling of storage.

Implements #33.

Test Plan:
----------
To manually test the delegate, shake the Simulator (^⌘Z) and toggle
"Set AsyncStorage Delegate". Otherwise, tests for JS have been duplicated for
the native delegate.

### DIFF
--- a/example/ios/AsyncStorageExample/AppDelegate.m
+++ b/example/ios/AsyncStorageExample/AppDelegate.m
@@ -7,13 +7,22 @@
 
 #import "AppDelegate.h"
 
+#import <RNCAsyncStorage/RNCAsyncStorage.h>
 #import <React/RCTBundleURLProvider.h>
+#import <React/RCTDevMenu.h>
 #import <React/RCTRootView.h>
 
-@implementation AppDelegate
+@implementation AppDelegate {
+  NSMutableDictionary<NSString *, NSString *> *_memoryStorage;
+}
 
 - (BOOL)application:(UIApplication *)application didFinishLaunchingWithOptions:(NSDictionary *)launchOptions
 {
+  [[NSNotificationCenter defaultCenter] addObserver:self
+                                           selector:@selector(didLoadJavaScript:)
+                                               name:RCTJavaScriptDidLoadNotification
+                                             object:nil];
+
   NSURL *jsCodeLocation;
 
   jsCodeLocation = [[RCTBundleURLProvider sharedSettings] jsBundleURLForBundleRoot:@"example/index" fallbackResource:nil];
@@ -30,6 +39,85 @@
   self.window.rootViewController = rootViewController;
   [self.window makeKeyAndVisible];
   return YES;
+}
+
+- (void)addDevMenuItemsForBridge:(RCTBridge *)bridge
+{
+  _memoryStorage = [NSMutableDictionary dictionary];
+
+  __weak AppDelegate *weakSelf = self;
+  RCTDevMenuItem *delegateToggle = [RCTDevMenuItem
+      buttonItemWithTitleBlock:^NSString * {
+        RNCAsyncStorage *asyncStorage = [bridge moduleForClass:[RNCAsyncStorage class]];
+        return asyncStorage.delegate == nil ? @"Set AsyncStorage Delegate"
+                                            : @"Unset AsyncStorage Delegate";
+      }
+      handler:^{
+        RNCAsyncStorage *asyncStorage = [bridge moduleForClass:[RNCAsyncStorage class]];
+        asyncStorage.delegate = asyncStorage.delegate == nil ? weakSelf : nil;
+      }];
+  [bridge.devMenu addItem:delegateToggle];
+}
+
+- (void)didLoadJavaScript:(NSNotification *)note
+{
+  RCTBridge *bridge = note.userInfo[@"bridge"];
+  if (bridge == nil) {
+    return;
+  }
+
+  [self addDevMenuItemsForBridge:bridge];
+}
+
+#pragma mark - RNCAsyncStorageDelegate
+
+- (void)allKeys:(nonnull RNCAsyncStorageResultCallback)completion
+{
+  completion(_memoryStorage.allKeys);
+}
+
+- (void)mergeValues:(nonnull NSArray<NSString *> *)values
+            forKeys:(nonnull NSArray<NSString *> *)keys
+         completion:(nonnull RNCAsyncStorageResultCallback)block
+{
+}
+
+- (void)removeAllValues:(nonnull RNCAsyncStorageCompletion)completion
+{
+  [_memoryStorage removeAllObjects];
+  completion(nil);
+}
+
+- (void)removeValuesForKeys:(nonnull NSArray<NSString *> *)keys
+                 completion:(nonnull RNCAsyncStorageResultCallback)completion
+{
+  for (NSString *key in keys) {
+    [_memoryStorage removeObjectForKey:key];
+  }
+  completion(@[]);
+}
+
+- (void)setValues:(nonnull NSArray<NSString *> *)values
+          forKeys:(nonnull NSArray<NSString *> *)keys
+       completion:(nonnull RNCAsyncStorageResultCallback)completion
+{
+  for (NSUInteger i = 0; i < values.count; ++i) {
+    NSString *value = values[i];
+    NSString *key = keys[i];
+    [_memoryStorage setObject:value forKey:key];
+  }
+  completion(@[]);
+}
+
+- (void)valuesForKeys:(nonnull NSArray<NSString *> *)keys
+           completion:(nonnull RNCAsyncStorageResultCallback)completion
+{
+  NSMutableArray *values = [NSMutableArray arrayWithCapacity:keys.count];
+  for (NSString *key in keys) {
+    NSString *value = _memoryStorage[key];
+    [values addObject:value == nil ? [NSNull null] : value];
+  }
+  completion(values);
 }
 
 @end
--- a/example/ios/AsyncStorageExample/AppDelegate.m
+++ b/example/ios/AsyncStorageExample/AppDelegate.m
@@ -14,6 +14,7 @@
 
 @implementation AppDelegate {
   NSMutableDictionary<NSString *, NSString *> *_memoryStorage;
+  __weak RCTBridge *_bridge;
 }
 
 - (BOOL)application:(UIApplication *)application didFinishLaunchingWithOptions:(NSDictionary *)launchOptions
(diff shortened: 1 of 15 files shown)


tido64:
@Krizzu: How are you testing RNCAsyncStorage? I had to patch `AsyncStorage.js` to get the correct implementation:
```diff
diff --git a/lib/AsyncStorage.js b/lib/AsyncStorage.js
index 4769097..6c007d8 100644
--- a/lib/AsyncStorage.js
+++ b/lib/AsyncStorage.js
@@ -18,6 +18,7 @@ const {NativeModules} = require('react-native');
 const RCTAsyncStorage =
   NativeModules.AsyncRocksDBStorage ||
   NativeModules.RNC_AsyncSQLiteDBStorage ||
+  NativeModules.RNCAsyncStorage ||
   NativeModules.AsyncLocalStorage;

 type ReadOnlyArrayString = $ReadOnlyArray<string>;
```

tido64:
@Krizzu: Also, it seems that in making it possible to run from Xcode, I broke Detox builds. Any suggestions here?

tido64:
Never mind. Figured out why Detox broke. 😁

tido64:
For the first question: #37

Krizzu:
@tido64 
This had to come up when we've changed naming on iOS side (some issues with camel case naming).

Your changes are already merged, so feel free to rebase.

tido64:
@Krizzu Thanks 👍

tido64:
@Krizzu Let me know how I can make this PR better 😉

Krizzu:
@tido64 Great, thanks!

I'm going to need a second pair of eyes on iOS code, but overall looks good. We could have some documentation about new feature - ideally in new `doc` file in `docs` directory. 

Also, some tests would be good to check it too.

ios/RNCAsyncStorageDelegate.h
```diff
@@ -0,0 +1,37 @@
...
+ */
+
+#import <Foundation/Foundation.h>
+
+NS_ASSUME_NONNULL_BEGIN
+
+typedef void (^RNCAsyncStorageCompletion)(NSError *_Nullable error);
+typedef void (^RNCAsyncStorageResultCallback)(NSArray<id<NSObject>> * valuesOrErrors);
+
+@protocol RNCAsyncStorageDelegate <NSObject>
```
zhongwuzw: @tido64 Thanks for work. I think it's necessary to support custom storage for users, currently if we set delegate, all operations would fall back to delegate. Ideally, I may like add an option to enable/disable built-in storage, the use case is, for example, if user only wants to observer the behavior of operations, not want to implement custom storage. @Krizzu What's your opinion? 🤔 .

tido64: FWIW, it should be trivial to add a property on the delegate whether to use it as a passthrough. If set to `YES`, we don't return and just continue as usual.

Krizzu: @tido64 	Sounds good to me. This would make AS more flexible for brownfield apps. @tido64 	What do you think?

tido64: I think this makes a lot of sense. There will be apps that want to monitor disk usage etc. but don't want to handle the storage themselves. I've just pushed the changes, let me know what you think. I'll work on docs/tests when I have time next.

ios/RNCAsyncStorageDelegate.h
```diff
@@ -0,0 +1,41 @@
...
+- (void)setValues:(NSArray<NSString *> *)values
+          forKeys:(NSArray<NSString *> *)keys
+       completion:(RNCAsyncStorageResultCallback)block;
+
+- (void)valuesForKeys:(NSArray<NSString *> *)keys
+           completion:(RNCAsyncStorageResultCallback)block;
+
+@optional
+
+@property (nonatomic, readonly, getter=isPassthrough) BOOL passthrough;
```
zhongwuzw: Please follow Camel-Case naming.

ios/RNCAsyncStorage.m
```diff
@@ -447,6 +556,16 @@ - (NSDictionary *)_writeEntry:(NSArray<NSString *> *)entry changedManifest:(BOOL
 
 RCT_EXPORT_METHOD(getAllKeys:(RCTResponseSenderBlock)callback)
 {
+  if (self.delegate != nil) {
+    [self.delegate allKeys:^(NSArray<id<NSObject>> *keys) {
+      callback(@[(id)kCFNull, keys]);
+    }];
+
+    if (![self _passthroughDelegate]) {
```
zhongwuzw: Is this opposite the meaning? I think if `passThrough` is `YES`, we should pass through built-in storage.

ios/RNCAsyncStorageDelegate.h
tido64: Not quite sure what you mean. [Passthrough](https://en.wikipedia.org/wiki/Passthrough) is one word.

ios/RNCAsyncStorage.m
tido64: If the delegate is a passthrough, it means that it wants to receive the event and also let the signal pass through it. So if it isn't, it will swallow the event, or return here.

Krizzu:
@tido64 	Sorry for conflict that arised  from merging #46.

ios/RNCAsyncStorageDelegate.h
zhongwuzw: Ahha, It's a word, sorry, but seems react-native uses `passThrough`, https://github.com/facebook/react-native/search?q=passThroughTouches&unscoped_q=passThroughTouches.

ios/RNCAsyncStorage.m
zhongwuzw: @tido64 Seems I misunderstand the `passthrough`.

zhongwuzw:
@Krizzu LGTM. I think we only need to add some tests and docs. 🤔

ios/RNCAsyncStorageDelegate.h
tido64: Yeah, it's a bit confusing but I think it's verb vs. noun. `passThroughTouches` returns whether touches should be passed down to child components, whilst here we're returning whether the delegate should be treated as a passthrough.

tido64:
@Krizzu @zhongwuzw I added docs and tests. Please have a look 😄

Krizzu:
@tido64 	Thanks! I'll have a look once I find a time later today

example/e2e/asyncstorage.spec.js
```diff
@@ -139,4 +152,84 @@ describe('Async Storage', () => {
...
+      await expect(element(by.id('testInput-shoe'))).toExist();
+    });
+
+    it('should crash when merging items in async storage', async () => {
+      if (device.getPlatform() === 'android') {
+        // Not yet supported.
+        return;
+      }
+
+      const buttonMergeItem = await element(by.id('mergeItem_button'));
```
Krizzu: Why failing on this test?

example/e2e/asyncstorage.spec.js
```diff
@@ -139,4 +152,84 @@ describe('Async Storage', () => {
...
+  });
+
+  describe('merge item delegate test', () => {
+    beforeAll(async () => {
+      if (device.getPlatform() === 'ios') {
+        await device.openURL({url: 'rnc-asyncstorage://set-delegate'});
+      }
+    });
+
+    it('should be visible', async () => {
```
Krizzu: This is already tested

example/e2e/asyncstorage.spec.js
```diff
@@ -139,4 +152,84 @@ describe('Async Storage', () => {
       expect(storyText).toHaveText(newStory);
     });
   });
+
+  describe('get / set / clear item delegate test', () => {
```
Krizzu: I feel like we should rather focus on testing if delegate is set/unset, otherwise we'd just repeated test that we already have.

example/ios/AsyncStorageExample/AppDelegate.m
```diff
@@ -32,4 +42,105 @@ - (BOOL)application:(UIApplication *)application didFinishLaunchingWithOptions:(
...
+    asyncStorage.delegate = self;
+  } else if ([url.host isEqualToString:@"unset-delegate"]) {
+    RNCAsyncStorage *asyncStorage = [_bridge moduleForClass:[RNCAsyncStorage class]];
+    asyncStorage.delegate = nil;
+  }
+
+  return YES;
+}
+
+#pragma mark - RNCAsyncStorageDelegate
```
Krizzu: Is it possible to move it to separate file?

example/e2e/asyncstorage.spec.js
```diff
@@ -29,12 +29,19 @@ describe('Async Storage', () => {
   });
 
   describe('get / set / clear item test', () => {
+    beforeAll(async () => {
+      if (device.getPlatform() === 'ios') {
+        await device.openURL({url: 'rnc-asyncstorage://unset-delegate'});
```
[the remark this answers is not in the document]
tido64: It is optional. I added these anyway to ensure that there are no inter-test dependencies and make them more robust. It shouldn't matter if somebody reorders them or the test runner shuffles the order.

example/e2e/asyncstorage.spec.js
tido64: Did you have anything special in mind? This change should be transparent from JS perspective. Would it make more sense to make all the delegate tests throw the keys/values it received, perhaps? That way we know for sure the delegate is handling the requests?

Krizzu: Sounds like a good idea - that way, we can be sure that the delegate is responding back, not default implementation.

tido64: Unfortunately, I couldn't get this to work reliably at all. Instead, I've rewritten the tests to expect different values coming from the delegate. For the get/set/clear tests, the delegate will always return the expected value + 1000000. For the merge test, "from delegate" string was appended to some fields. Please take a look 😄

stigi:
Hi team,

just a heads up that this breaks native linking via cocoapods, as the library is linked as `react-native-async-storage` and expects imports of headers as `#import <react-native-async-storage/RNCAsyncStorageDelegate.h>` and not `#import <RNCAsyncStorage/RNCAsyncStorageDelegate.h>`.

There are two ways to move forward:
- change the import to `#import <react-native-async-storage/RNCAsyncStorageDelegate.h>`
- rename the pod (name of .podspec file and value of `s.name`) to `RNCAsyncStorage`

Let me know, I'd be happy to put up a PR

Krizzu:
@stigi Thanks for heads up!

I'm happy to go with the first approach. WDYT @tido64 ?

stigi:
The second approach also requires the current user of cocoa pod linking to update their Podfiles, so option 1 is less intrusive, even though crunch seeing the import from `<react-native-async-storage...>` :neckbeard:

tido64:
Sorry for testing this thoroughly with Pods. Completely slipped my mind.

I’m for solution #2. Yes, it will break on upgrade but it’s the right thing to do and will be in line with Apple’s naming guidelines and what the community will expect.

Thanks for bringing it to our attention and fixing it ❤️

lightman73:
First, thank you, you just saved me. I spent the last hour trying to understand why the hell Xcode wasn't finding RNCAsyncStorageDelegate.h ... 

I manually implemented solution #1 and it's building everything as expected. 

That said, I vote for #2 as well. Although it is breaking on upgrade, it's the cleanest solution, imho. 

Thank again !

Krizzu:
Okay, seems like we made a decision.
Just to be clear: what steps would have to be taken to successfully upgrade with solution 2?

tido64:
Somebody correct me if I'm mistaken, but if you're using `react-native link`, isn't it just a matter of re-linking it? We use manual linking at work so we would just fix the relevant line in our `Podfile`.

tido64:
@Krizzu @stigi Sorry, I got a little impatient since I broke this release. I've implemented the fix here: #75. I've checked that `react-native link` will pick up the new `.podspec` but it won't clean up the old one. We will unfortunately have to ask users to go into their `Podfile` and make manual changes. 😞

Issue #51, opened 2019-03-29
Linking when using kotlin not possible? NativeModule.RCTAsyncStorage is null.


rpuls:
## You want to:  
Since I was getting warnings that the react-native asyncStorags is becomming deprecated and is moved (here) to individual package, I decided to install... This seems to be a pain in the a** every time, was also troubleful when I installed RCs' extracted WebView. (grumpy sorry)....

Anyway, this time I am stuck with failing jest tests that I have not been able to resolve after 24 hours of trying. Have had multiple colleges passing by as well as asked for help on SO.
Please check this link for further issue despription: https://stackoverflow.com/questions/55418910/jest-test-fails-after-installing-react-native-async-storage

I did manual linking since `react-native link` will not work for this module as MainApplication.java does not exist when using kotlin. Instead I have MainApplication.kt

After linking manually the async functions are working just fine. Except the jest test are unable to find the module. 
` @RNCommunity/AsyncStorage: NativeModule.RCTAsyncStorage is null.`

Any suggestions?

tbergq:
Hi, I answered your post on stackoverflow. 

Also see [our repo](https://github.com/kiwicom/mobile/blob/master/app/core/config/NativeModulesMocks/AsyncStorage.js) if you like.

Krizzu:
@rpuls 	Hey, thanks for reaching out.

I'm on the way of writing mocks to be used with Jest. I'm aiming to finish in few days, alongside with docs explaining how to test it.

For now, to get by this error, you'd have to mock `NativeModule` from `react-native`, to return [either of those modules](https://github.com/react-native-community/react-native-async-storage/blob/master/lib/AsyncStorage.js#L17).

rpuls:
@Krizzu 
> I'm on the way of writing mocks to be used with Jest. I'm aiming to finish in few days, alongside with docs explaining how to test it.
> 
> For now, to get by this error, you'd have to mock `NativeModule` from `react-native`, to return [either of those modules](https://github.com/react-native-community/react-native-async-storage/blob/master/lib/AsyncStorage.js#L17).

Hey thanks for replying! I found a solution that worked. Tried many different ways of mocking the AsyncStorage. 4th attempt worked like a charm. I updated my stackoverflow with the solution https://stackoverflow.com/questions/55418910/jest-test-fails-after-installing-react-native-async-storage/55421364?noredirect=1#comment97619261_55421364

Krizzu:
Hey @rpuls,

[...]

Issue #39, opened 2019-03-14
SyntaxError: Unexpected token export using jest


loic-lopez:
## Current behavior

```bash
/@react-native-community/async-storage/lib/index.js:1
    ({"Object.<anonymous>":function(module,exports,require,__dirname,__filename,global,jest){export default from './AsyncStorage';

 SyntaxError: Unexpected token export

    > 1 | import AsyncStorage  from "@react-native-community/async-storage"
        | ^
      2 | 
      3 | /**
      4 |  * Loads a string from storage.

```

## Repro steps

`import @react-native-community/async-storage` and test with jest.

## Environment

- **Async Storage version**: 1.2.1
- **React-Native version**: 0.57.7
- **Platform tested**: iOS / Android
- **Jest version**: 24.0.11
- **Logs/Error that are relevant**: see above

Krizzu:
Hey @loic-lopez 

Async Storage is not being transformed by Jest, so in your `jest.config.js` or `jest` entry in `package.json`, add:

```
"transformIgnorePatterns": ["/node_modules/@react-native-community/async-storage/(?!(lib))"]
```

vgm8:
The error still happening after adding your solution.

loic-lopez:
@Krizzu thanks for reply and yes `transformIgnorePatterns` seems to have solved the problem but,
when i try to mock the library like that:
<img width="508" alt="Capture d’écran 2019-03-15 à 11 30 54" src="https://user-images.githubusercontent.com/17724398/54442869-d8481a00-4715-11e9-9282-bbf42b52ad8a.png">

**I got:** 
```bash
● save

    expect(jest.fn()).toHaveBeenCalledWith(expected)

    Expected mock function to have been called with:
      ["something", "{\"x\":1}"]
    But it was not called.

      34 | test("save", async () => {
      35 |   await save("something", VALUE_OBJECT)
    > 36 |   expect(mockSetItem).toHaveBeenCalledWith("something", VALUE_STRING)
         |                       ^
      37 | })
      38 | 
      39 | test("saveString", async () => {

      at Object.<anonymous>.test (test/tests/storage.test.ts:36:23)
```

**as if the mock did not work**

Krizzu:
@loic-lopez This is more like Jest question :) You can checkout [Jest docs about mocking node_modules libs](https://jestjs.io/docs/en/manual-mocks#mocking-node-modules).

[...]

Issue #12, opened 2019-02-15
AsyncStorage.getItem() doesn't seem to work


cpojer:
* * *
  This issue was originally created by @mrded as facebook/react-native#18372.
  * * *

  

When I try to set a value via `AsyncStorage.getItem()`, I cannot request it back.

### Environment

Environment:
  OS: macOS High Sierra 10.13.3
  Node: 9.8.0
  Yarn: 1.5.1
  npm: 5.6.0
  Watchman: 4.9.0
  Xcode: Xcode 9.2 Build version 9C40b
  Android Studio: Not Found

Packages: (wanted => installed)
  react: ^16.3.0-alpha.1 => 16.3.0-alpha.1
  react-native: 0.54.0 => 0.54.0

### Expected Behavior

    await AsyncStorage.setItem('foo', 'bar');
    await AsyncStorage.getItem('foo').then(console.log); // 'bar'
    await AsyncStorage.getAllKeys().then(console.log); // ['foo']

### Actual Behavior

    await AsyncStorage.setItem('foo', 'bar');
    await AsyncStorage.getItem('foo').then(console.log); // null
    await AsyncStorage.getAllKeys().then(console.log); // []

### Steps to Reproduce

    import { AsyncStorage } from 'react-native';

    it('should be able to request a value after it was set', async () => {
      await AsyncStorage.setItem('foo', 'bar');
      const output = await AsyncStorage.getItem('foo');

      expect(output).toBe('bar');
    });

    it('should be able to see a new key after a value was set', async () => {
      await AsyncStorage.setItem('foo', 'bar');
      const output = await AsyncStorage.getAllKeys();

      expect(output).toContain('foo');
    });

joshjhargreaves:
We recently saw issues with this in our app on a large number of android devices when doing something very similar to the above in production but we had reproducing locally. Very happy to help do the work necessary to fix this. Would be great to use this thread to maybe identify what's causing the issue and different approaches we could use to fix it!

retyui:
Maybe an author get null because use of mock implementation:
https://github.com/facebook/react-native/blob/f741d338406d518c4f81fada723991b2938b8cc5/jest/setup.js#L77-L92

Krizzu:
Hey @joshjhargreaves 	@retyui 	

Thanks for your time. 
To dig deeper into the issue, I think we'd need more info about reproducing it - from repro steps, it seems like the problem is happening in a test environment (like @retyui pointed out).

[...]

PR #53, opened 2019-04-01
[feat] Jest mocks, AsyncStorage testing guidelines

Krizzu:
Summary:
---------

We had some issues with Async Storage when it comes to testing with Jest. This PR adds mocks to be used while testing, to avoid error and integrate AS better.

Also some updates to README/Error thrown when module is not linked.

Small fix for CI job.

This PR will close:
#51, #39, #12

Test Plan:
----------

Green CI, test it with an app.


docs/Jest-integration.md
```diff
@@ -0,0 +1,100 @@
...
+1. In your project root directory, create `__mocks__` directory.
+2. Inside that folder, create `react-native.js` file.
+3. Inside that file, import `AsyncStorageMock` and use it to mock `NativeModules`.
+
+
+```js
+
+import AsyncStorageMock from '@react-native-community/async-storage/jest/async-storage-mock';
+
+jest.doMock('react-native', () => {
```
thymikee: I think it's better to use `jest.mock`, because it's more widely used

docs/Jest-integration.md
```diff
@@ -0,0 +1,100 @@
...
+
+You can use mocked Async Storage in one of two ways:
+
+### Mock `node_modules`
+
+1. In your project root directory, create `__mocks__` directory.
+2. Inside that folder, create `react-native.js` file.
+3. Inside that file, import `AsyncStorageMock` and use it to mock `NativeModules`.
+
+
```
thymikee: FYI you can run Prettier on .md files to remove excess whitespace :)

docs/Jest-integration.md
```diff
@@ -0,0 +1,100 @@
...
+1. In Jest config (probably in `package.json`) add setup files location:
+
+```json
+// ...
+  "jest": {
+    // other config
+    "setupFiles": "./path/to/jestSetupFile.js"
+  },
+// ...
```
thymikee: This is not a valid JSON. Also `setupFiles` accepts an array. Please fix to: 
```json
{
  "jest": {
    "setupFiles": ["./path/to/jestSetupFile.js"]
  }
}
```

docs/Jest-integration.md
```diff
@@ -0,0 +1,100 @@
...
+
+
+1. **`SyntaxError: Unexpected token export` in async-storage/lib/index.js**
+
+This is likely because `Jest` is not transforming Async Storage. You can point it to do so, by adding line in Jest's configuration:
+
+```json
+  "jest": {
+    // ...
+    "transformIgnorePatterns": ["/node_modules/@react-native-community/async-storage/(?!(lib))"]
```
thymikee: 1. Make it valid JSON
2. Use a catch-all regex for all `@react-native-community` packages to help others :D :
```
"transformIgnorePatterns": ["node_modules/(?!react-native|@react-native-community)"]
```
3. Make a note that this should be included by default in RN 0.60

jest/async-storage-mock.js
```diff
@@ -0,0 +1,50 @@
...
+
+type GetKeysType = (error: ?Error, keys: KeysType) => void;
+
+type MultiGetType = (error: ?Error, result: KeyValueType) => void;
+
+const AsyncStorageMock = {
+  /**
+   * Clears all keys in storage
+   */
+  clear: (cb: CallbackType) => cb(),
```
thymikee: So the mock is actually test-framework agnostic. Maybe it's nice, but I think most of the users would expect this mock to be fully functional in JS-land only, plus it would be nice if it used `jest.fn` to mock methods, because that would make them, well, mocks :P which you can inspect.

How about getting rid of the example mock and just provide the basic implementation + mocks here?

lib/AsyncStorage.js
```diff
@@ -17,13 +17,14 @@ const RCTAsyncStorage =
...
 if (!RCTAsyncStorage) {
-  throw new Error(`@RNCommunity/AsyncStorage: NativeModule.RCTAsyncStorage is null.
+  throw new Error(`[@RNC/AsyncStorage]: NativeModule: AsyncStorage is null.
 
 To fix this issue try these steps:
   • Run \`react-native link @react-native-community/async-storage\` in the project root.
   • Rebuild and re-run the app.
   • Restart the packager with \`--clearCache\` flag.
   • If you are using CocoaPods on iOS, run \`pod install\` in the \`ios\` directory and then rebuild and re-run the app.
+  • If this error shows up while testing with Jest, check out how to mock Async Storage in repository docs.
```
thymikee: it would be nice to provide a link

package.json
```diff
@@ -54,7 +54,8 @@
     "semantic-release": "15.13.3"
   },
   "jest": {
-    "preset": "react-native"
+    "preset": "react-native",
+    "testMatch": [ "**/__tests__/**/*.[jt]s?(x)", "**/?(*.)+(test).[jt]s?(x)" ]
```
thymikee: this looks pretty much lika a default testMatch, do you really need to change it?. Also, format with Prettier

jest/async-storage-mock.js
Krizzu: That was my dilemma here - to provide fully functional mock or simple one + docs about providing your own functionality. I decided to do the latter, as I thought most people would use it to pass tests. 

But I think you might be right here - it's much easier for users to implement it other way around. Going to make it fully functional first.

Also, I had `jest.fn` at first, but then I thought about accessing this mock in your test - how would you do that? You'd have to import mock into your test and check, i.e if it was called with such arguments. Or am I missing something here?

thymikee: If you substitute module with a mock like `RN.AsyncStorage = ASMock;`, `AsyncStorage` should be a mock then and you'll be able to do `expect(AsyncStorage.getItem).toBeCalledWith(smth)`

package.json
Krizzu: Good point, I left `.spec` out, for e2e tests. Going to change that.

jest/async-storage-mock.js
Krizzu: Yeah, but this is mock for `NativeModule`, not something that user has access too

Krizzu:
@thymikee 
This is rather large change from what I had before - now instead mocking `NativeModule`, I simply mock JS interface itself. 

All of public facing methods of the mock is not `jest.fn`, meaning more test cases can be checked. 

Also, added unit test to test mock itself, for both Promise and Callback functionality.


Thanks for your review so far and please let me know what you think 🙏

docs/Jest-integration.md
```diff
@@ -0,0 +1,77 @@
+# Jest integration
+
+Async Storage module is tighly coupled with a `Native Module`, meaning it needs a running React Native application to work properly. In order to use it in tests, you have to provide its separate implementation. Follow those steps to add mocked `Async Storage` to your test cases.
```
thymikee: `Native Module` -> `NativeModule`

docs/Jest-integration.md
```diff
@@ -0,0 +1,77 @@
...
+  import mockAsyncStorage from '@react-native-community/async-storage/jest/async-storage-mock';
+
+  jest.mock('@react-native-community/async-storage', () => mockAsyncStorage);
+```
+## Testing with mock
+
+Each public method available from `Async Storage` is [a mock function](https://jestjs.io/docs/en/mock-functions), that you can test for certain condition, for example, check if it has been called with a specific arguments:
+
+```javascript
+  it('checks if Async Storage is used', async () => {
```
thymikee: wonder why the indentation? Are you using Prettier to format that? If not, I definitely recommend

docs/Jest-integration.md
thymikee: I don't see this addressed, mind fixing?

example/__tests__/App.js
```diff
@@ -1,15 +1,157 @@
...
+              expect(data2).not.toBeNull();
+
+              AsyncStorage.removeItem('random1', () => {
+                AsyncStorage.removeItem('random2', () => {
+                  AsyncStorage.getItem('random1', (_, value1) => {
+                    AsyncStorage.getItem('random2', (_, value2) => {
+                      expect(value1).toBeNull();
+                      expect(value2).toBeNull();
+                      done();
+                    }).catch(e => done.fail(e));
```
thymikee: you're using callback approach, but here taking advantage of `getItem` returning a promise. You should use the first argument (named `_` instead of `err` or `error`) and check if it's not null, then call `done.fail`

Krizzu: The reason why I'm catching some of the block is that `error` thrown in callback methods are swallowed by Jest - meaning it'll get `timeout` singal, without specific reason why caused that.

Note that I only wrap blocks with assertions inside - I couldn't find nicer way to do `callbacks` and still get errors when assertion fail

docs/Jest-integration.md
```diff
@@ -0,0 +1,79 @@
...
+### **`SyntaxError: Unexpected token export` in async-storage/lib/index.js**
+
+**Note:** In React Native 0.60+, all `@react-native-community` packages are transformed by default.
+
+You need to point Jest to transform this package. You can do so, by adding `transformIgnorePatterns` setting in Jest's configuration. This path includes all RN Community packages:
+
+
+```json
+"jest": {
+  "transformIgnorePatterns": ["node_modules/(?!(@react-native-community))/"]
```
thymikee:
```suggestion
  "transformIgnorePatterns": ["node_modules/(?!(react-native|@react-native-community))"]
```

PR #54, opened 2019-04-03
fix: Fix Windows support

sbeca:
Summary:
---------

The [react-native-windows](https://github.com/Microsoft/react-native-windows) repo provides Windows build target support to React Native projects. It provides AsyncStorage support through the module name `AsyncLocalStorage` (exact file is [here](https://github.com/Microsoft/react-native-windows/blob/master/RNWCS/ReactWindows/ReactNative/Modules/Storage/AsyncStorageModule.cs)) but commit 547445c148b18be8bd1476d418e6f3318587b5cb removed the reference to `NativeModules.AsyncLocalStorage` from this repo which broke Windows support. This PR adds that back in.

Test Plan:
----------

Green CI

### DIFF
--- a/lib/AsyncStorage.js
+++ b/lib/AsyncStorage.js
@@ -14,7 +14,9 @@
 const {NativeModules} = require('react-native');
 
 const RCTAsyncStorage =
-  NativeModules.RNC_AsyncSQLiteDBStorage || NativeModules.RNCAsyncStorage;
+  NativeModules.RNC_AsyncSQLiteDBStorage ||
+  NativeModules.RNCAsyncStorage ||
+  NativeModules.AsyncLocalStorage; // NativeModules.AsyncLocalStorage provides compatibility with out out tree platform
 
 if (!RCTAsyncStorage) {
   throw new Error(`@RNCommunity/AsyncStorage: NativeModule.RCTAsyncStorage is null.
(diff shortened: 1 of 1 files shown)


lib/AsyncStorage.js
```diff
@@ -14,7 +14,9 @@
 const {NativeModules} = require('react-native');
 
 const RCTAsyncStorage =
-  NativeModules.RNC_AsyncSQLiteDBStorage || NativeModules.RNCAsyncStorage;
+  NativeModules.RNC_AsyncSQLiteDBStorage ||
+  NativeModules.RNCAsyncStorage ||
+  NativeModules.AsyncLocalStorage; // NativeModules.AsyncLocalStorage provides compatibility with out out tree platform
```
tido64: Could we move this to the top? That would allow custom implementation on "in-tree" platforms.

Krizzu:
Hey @sbeca, thanks for bringing this up!

I've deleted this line as it looked like old-code leftover. I'm happy to bring it back.

lib/AsyncStorage.js
Krizzu: I don't quite get what "tree platform" is. I don't see a problem with moving it on top.

Krizzu:
```suggestion
  NativeModules.AsyncLocalStorage; // Support for external modules, like react-native-windows
```

sbeca:
Thank you for the review @tido64 and @Krizzu. I have just pushed a new commit that includes both of your recommendations.

PR #57, opened 2019-04-04
React Native peer dependency warning

mifi:
Summary:
---------

`yarn add @react-native-community/async-storage`
when running rn 0.59.0 gives:

warning " > @react-native-community/async-storage@1.2.2" has incorrect peer dependency "react-native@^0.57.0".

Test Plan:
----------

### DIFF
--- a/package.json
+++ b/package.json
@@ -34,7 +34,7 @@
   },
   "peerDependencies": {
     "react": "^16.0",
-    "react-native": "^0.59.0"
+    "react-native": ">=0.57"
   },
   "devDependencies": {
     "@babel/core": "7.0.0",
--- a/package.json
+++ b/package.json
@@ -34,7 +34,7 @@
   },
   "peerDependencies": {
     "react": "^16.0",
-    "react-native": "^0.57.0"
+    "react-native": "^0.59.0"
   },
   "devDependencies": {
     "@babel/core": "7.0.0",


package.json
```diff
@@ -34,7 +34,7 @@
   },
   "peerDependencies": {
     "react": "^16.0",
-    "react-native": "^0.57.0"
+    "react-native": "^0.59.0"
```
tido64: Would it make more sense to have something like `>=0.57.0 <1.0.0`? I'm not comfortable with adding a warning on users still on older versions. And what about future ones?

guhungry: I think * would be fine too since async storage doesn't depend any react functionality but bridging between js and native module.

mifi: Isn't it using some kind of API provided by react native at all? like the bridge api could probably change, right @guhungry ?
@tido64 yes, but what if RN makes a breaking change in the future, then it will report incorrectly that it is compatible. I think it should be from version to version, but i don't know if that is possible?

tido64: Then I'd suggest at least `>=0.57.0 <0.60.0`. Native modules probably isn't going to be removed within the next few releases.

Krizzu: @mifi Yup, bridge API will change in the future.
Let's stick to `>=0.57`, as those version are tested with Async Storage, and when breaking change will appear in upcoming releases, we'll add necessary adjustments.

Krizzu:
Hey @mifi ,

Can you provide a status of this PR? 


thanks.

mifi:
I updated the pr now

PR #63, opened 2019-04-08
fix: Prefer RNC's AsyncStorage over RN's.

tido64:
Summary:
---------
React Native's (deprecated) AsyncLocalStorage is being picked up before RNC's.

Test Plan:
----------
All the current tests should still pass.


tido64:
cc @Krizzu

PR #64, opened 2019-04-09
iOS clean migration from RNCAsyncStorage_V1 to RTCAsyncStorage_V1

reilem:
Summary:
---------

See: https://github.com/react-native-community/react-native-async-storage/issues/55

Test Plan:
----------

The only type of tests I have already run is:
1. Hardcode set the variable `RCTStorageDirectory` to `RNCAsyncStorage_V1`
2. Run the app & set some values to cache
3. Close the app
4. Reset `RCTStorageDirectory` to `RCTAsyncStorage_V1`
5. Run the app and see that cached values have been migrated

I have performed these tests with and without a pre-existing storage directory and they work in both cases.

My only concerns are the way I am currently handling errors. If something goes wrong an `NSException` is thrown which will trigger a SIGABRT shutdown of the app since it is never caught. I am not extremely experienced with error handling in Objective-C so perhaps someone has some better ideas.

### DIFF
--- a/ios/RNCAsyncStorage.m
+++ b/ios/RNCAsyncStorage.m
@@ -101,7 +101,7 @@ static NSString *RCTGetStorageDirectory()
 
 static NSString *RCTCreateManifestFilePath(NSString *storageDirectory)
 {
-  return [RCTCreateStorageDirectoryPath(storageDirectory) stringByAppendingString:RCTManifestFileName];
+  return [RCTCreateStorageDirectoryPath(storageDirectory) stringByAppendingPathComponent:RCTManifestFileName];
 }
 
 static NSString *RCTGetManifestFilePath()
--- a/ios/RNCAsyncStorage.m
+++ b/ios/RNCAsyncStorage.m
@@ -506,7 +506,7 @@ RCT_EXPORT_METHOD(multiSet:(NSArray<NSArray<NSString *> *> *)kvPairs
 }
 
 RCT_EXPORT_METHOD(multiMerge:(NSArray<NSArray<NSString *> *> *)kvPairs
-                  callback:(RCTResponseSenderBlock)callback)
+                    callback:(RCTResponseSenderBlock)callback)
 {
   if (self.delegate != nil) {
     NSMutableArray<NSString *> *keys = [NSMutableArray arrayWithCapacity:kvPairs.count];
@@ -640,4 +640,3 @@ RCT_EXPORT_METHOD(getAllKeys:(RCTResponseSenderBlock)callback)
 }
 
 @end
-
--- a/ios/RNCAsyncStorage.m
+++ b/ios/RNCAsyncStorage.m
@@ -17,6 +17,7 @@
 #import <React/RCTUtils.h>
 
 static NSString *const RCTStorageDirectory = @"RCTAsyncLocalStorage_V1";
+static NSString *const RCTOldStorageDirectory = @"RNCAsyncLocalStorage_V1";
 static NSString *const RCTManifestFileName = @"manifest.json";
 static const NSUInteger RCTInlineValueThreshold = 1024;
 
@@ -185,7 +186,6 @@ static NSDate *RCTManifestModificationDate(NSString *manifestFilePath)
   return [attributes fileModificationDate];
 }
 
-static NSString *const RCTOldStorageDirectory = @"RNCAsyncLocalStorage_V1";
 /**
  * Creates an NSException used during Storage Directory Migration.
  */
--- a/ios/RNCAsyncStorage.m
+++ b/ios/RNCAsyncStorage.m
@@ -174,61 +174,50 @@ static NSDictionary *RCTDeleteStorageDirectory()
   return error ? RCTMakeError(@"Failed to delete storage directory.", error, nil) : nil;
 }
 
+
+NSString *OldStorageDirectory = @"RNCAsyncLocalStorage_V1";
+/**
+ * Creates an NSException used during Storage Directory Migration.
+ */
+static NSException *RCTStorageDirectionMigrationException(NSString *reason, NSError *error) {
+  NSMutableString *reasonString = [[NSMutableString alloc] initWithString:reason];
+  [reasonString appendString:@" - "];
+  [reasonString appendString:[error description]];
+  return [[NSException alloc] initWithName:@"RCTStorageDirectoryMigrationFailure" reason:reasonString userInfo:nil];
+}
+
 /**
  * This check is added to make sure that anyone coming from pre-1.2.2 does not lose cached data.
  * Data is migrated from the "RNCAsyncLocalStorage_V1" directory to the "RCTAsyncLocalStorage_V1" directory.
  */
 static void RCTStorageDirectoryMigrationCheck() {
-  NSString *oldStorageDir = RCTCreateStorageDirectoryPath(@"RNCAsyncLocalStorage_V1");
-
-  NSError *error;
-  BOOL isDir;
-  BOOL oldExists = [[NSFileManager defaultManager] fileExistsAtPath:oldStorageDir isDirectory:&isDir];
-
-  // If the old directory exists, it means we need to migrate data to the new directory
-  if (oldExists && isDir) {
-
-    BOOL migrationSuccess;
-    // If the new storage directory already exists, then this may be caused by ever older data being left behind from previous versions.
-    // This old data will need to be overwritten.
-    NSString *newStorageDir = RCTGetStorageDirectory();
-    BOOL newExists = [[NSFileManager defaultManager] fileExistsAtPath:newStorageDir isDirectory:&isDir];
-    if (newExists && isDir) {
-      NSString *backupStorageDir = RCTCreateStorageDirectoryPath(@"RCTBackupAsyncLocalStorage_V1");
-      // Replace any possible existing data in the new storage with old storage directory
-      migrationSuccess = [[NSFileManager defaultManager] replaceItemAtURL:[[NSURL alloc] initWithString:newStorageDir]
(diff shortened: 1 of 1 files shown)


Krizzu:
Hey @reilem, thanks for your work!
Code looks fine to me, so let me just test this on my end and let's merge it afterwards.

ios/RNCAsyncStorage.m
```diff
@@ -156,6 +162,52 @@ static dispatch_queue_t RCTGetMethodQueue()
   return error ? RCTMakeError(@"Failed to delete storage directory.", error, nil) : nil;
 }
 
+
+NSString *OldStorageDirectory = @"RNCAsyncLocalStorage_V1";
```
tido64: Can you add `const` here to prevent accidental overwrite? Also, add a prefix since this lives in the global namespace, and move it to the top along with the other constants.
```objc
NSString *const RCTOldStorageDirectory = @"RNCAsyncLocalStorage_V1";
```

ios/RNCAsyncStorage.m
```diff
@@ -156,6 +162,52 @@ static dispatch_queue_t RCTGetMethodQueue()
...
+/**
+ * Creates an NSException used during Storage Directory Migration.
+ */
+static NSException *RCTStorageDirectionMigrationException(NSString *reason, NSError *error) {
+  NSMutableString *reasonString = [[NSMutableString alloc] initWithString:reason];
```
tido64: Instead of mutating a string, we can use `stringWithFormat:`:
```objc
NSString* detailedReason = [NSString stringWithFormat:@"%@: %@", reason, error.description];
```

ios/RNCAsyncStorage.m
```diff
@@ -156,6 +162,52 @@ static dispatch_queue_t RCTGetMethodQueue()
...
+  dispatch_once(&onceToken, ^{
+    NSError *error;
+    BOOL isDir;
+    // If the old directory exists, it means we need to migrate data to the new directory
+    if ([[NSFileManager defaultManager] fileExistsAtPath:RCTCreateStorageDirectoryPath(OldStorageDirectory) isDirectory:&isDir] && isDir) {
+      // Check if the new storage directory location already exists
+      BOOL newStorageDirectoryExists = [[NSFileManager defaultManager] fileExistsAtPath:RCTGetStorageDirectory() isDirectory:&isDir];
+      if (newStorageDirectoryExists) {
+        // If the new storage directory location already exists, remove existing directory
+        newStorageDirectoryExists = !([[NSFileManager defaultManager] removeItemAtPath:RCTGetStorageDirectory() error:&error]);
```
tido64: I can't help but think that we might lose data here depending on which version you're migrating from. Would it be better to check for last modified folder and migrate that one? I think the `manifest.json` file is updated with every write and can be used as a pretty good indicator.

ios/RNCAsyncStorage.m
```diff
@@ -167,6 +219,19 @@ @implementation RNCAsyncStorage
   NSMutableDictionary<NSString *, NSString *> *_manifest;
 }
 
++ (BOOL)requiresMainQueueSetup {
+  return YES;
```
tido64: Why do we require main queue? Isn't `NSFileManager` thread safe?

ios/RNCAsyncStorage.m
```diff
@@ -156,6 +162,52 @@ static dispatch_queue_t RCTGetMethodQueue()
...
+        // If the new storage directory location already exists, remove existing directory
+        newStorageDirectoryExists = !([[NSFileManager defaultManager] removeItemAtPath:RCTGetStorageDirectory() error:&error]);
+        if (newStorageDirectoryExists) {
+          @throw RCTStorageDirectionMigrationException(@"Failed to clear pre-existing storage directory", error);
+        }
+      }
+      if (!newStorageDirectoryExists) {
+        // If new storage direction doesn't exist, copy old storage directory to new location
+        if (![[NSFileManager defaultManager] copyItemAtPath:RCTCreateStorageDirectoryPath(OldStorageDirectory) toPath:RCTGetStorageDirectory() error:&error]) {
+          @throw RCTStorageDirectionMigrationException(@"Failed to copy old storage directory to new storage directory", error);
```
tido64: We probably shouldn't take down the whole app if we fail to migrate. I think in the rare cases where IO might fail, maybe it's probably fine to just say that we "lost" the data?

Krizzu:
@tido64 Thanks for having a look. @reilem Thank you for your work 💪 

On a second though, simply deleting `RCT` folder first would lead to some data loss as well. 

Say, someone used pre-1.2.2 AS, then upgrades to 1.2.3/4 (`RCT` used here, meaning "fresh start"). Now, upgrading to next version (with this feature included), will again end with overriding with stuff from `RNC`.

I like the approach that @tido64 suggested, to check when each of the files (from RNC and RTC) were accessed and based on that proceed with migration.

Something like that:

```
if(RNC_DIR.exists){
    if(RCT_DIR.exists){
        if(RNC_DIR.accessTime > RCT_DIR.accessTime){
            // RNC being used, pre 1.2.2
            rm RCT_DIR;
            mv RNC_DIR RCT_DIR;
        else {
            // RCT in use, so ditch RNC
            rm RNC_DIR;
        }
    else {
        // RCT does not exists, so move RNC
        mv RNC_DIR RCT_DIR;
    }
} 

```

Let me know what you think!

ios/RNCAsyncStorage.m
reilem: Ah thanks, I was looking for something like this. 😄

ios/RNCAsyncStorage.m
reilem: This does sound a lot safer, I'll have a look at the manifest file.

ios/RNCAsyncStorage.m
reilem: Indeed, I looked into it and I think I can return `NO` here since we are not using any UIKit elements. React-native tends to complain if you `init` without implementing `requiresMainQueueSetup` and I guess my initial thought was to return `YES` just to be sure.

On an unrelated note, NSFileManager is thread safe unless you "use a delegate to receive notifications about the status of move, copy, remove, and link operations". In which case using a shared `defaultManager` is not so safe. But I don't think that's being done here. :) https://developer.apple.com/documentation/foundation/nsfilemanager

ios/RNCAsyncStorage.m
reilem: Yeah I am not a big fan of crashing the app either, but I'm not a fan of leaving errors unused either. I guess in this case it would be just fine to ignore the errors? Cause I don't think adding `NSLog`s will be of much use either.

tido64: I think there's also a real risk of going into a crash loop. I think we should ignore it, and use [RCTLogWarn or RCTLogError](https://github.com/facebook/react-native/blob/629708bedae65a30e39d234da6b04d6fa101a779/React/Base/RCTLog.h) to log errors. What do you think, @Krizzu?

reilem: Ah cool, didn't know these existed. Errors would also cause app crashes so I guess we could go with the Warnings to be safe. Right now I've replaced everything with NSLogs for now, so let me know what you guys think is best.

reilem:
I've made the requested changes however I have encountered a strange issue during testing. While the `NSFileManager` displays that old storage directories have been successfully removed and no longer exist (explicitly checked) I still somehow manage to get the actual data displayed in my app. I'm guessing that my form of testing really isn't robust enough since it seems like some sort of cached data is being accessed.

Aside from the above issue, I have debugged the flow of the migration in the four cases and they seem to be working:
1. Old data not exists --> do nothing
2. Old data exists --> New data not exists --> copy --> remove old
3. Old data exists --> New data exists --> New data is newer --> remove old
4. Old data exists --> New data exists --> Old data is newer --> remove new --> copy --> remove old

Krizzu:
@reilem Hey,

This strange behavior might come from the regression we had on iOS in version 1.2.3 and 1.2.4. Please rebase with latest master and check if it works for you and let me know.

ios/RNCAsyncStorage.m
tido64: I think I prefer `RCTLogWarn` because they will at least be visible during debug as a yellow box, in addition to being logged to the console.

reilem:
I think I discovered my issue, during refactoring I made the manifest file path using `stringByAppendingString` instead of `stringByAppendingPathComponent`.

reilem:
I have rebased on the latest PR merge, fixed the manifest issue and now use `RCTLogWarn`. All my "manual" tests are working now without issue. I thought about automating the tests, but I don't know of any accurate way to automate app restarts and I couldn't find anything about mocking const values in objective-c. Unless you guys have any ideas? 😄

Krizzu:
Looking good! Going to test it manually later today to confirm it from my side.

thanks.

ios/RNCAsyncStorage.m
```diff
@@ -168,6 +179,75 @@ static dispatch_queue_t RCTGetMethodQueue()
   return error ? RCTMakeError(@"Failed to delete storage directory.", error, nil) : nil;
 }
 
+static NSDate *RCTManifestModificationDate(NSString *manifestFilePath)
+{
+  NSDictionary *attributes = [[NSFileManager defaultManager] attributesOfItemAtPath:manifestFilePath error:nil];
+  return [attributes fileModificationDate];
+}
+
+static NSString *const RCTOldStorageDirectory = @"RNCAsyncLocalStorage_V1";
```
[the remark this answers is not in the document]
reilem: Done 😄

Krizzu:
@reilem @tido64 Thanks guys! 🥇

PR #74, opened 2019-04-12
Fix #73

### LINKED ISSUE #73: Recent 1.3.1 release breaks iOS build if linked with pods
Screenshot 2019-04-12 14 45 19

## Repro steps

I've created an example app at https://github.com/timmywil/react-native-async-storage-1.3.1

This app was created with `react-native init` and async-storage included `react-native link... && cd ios/ && pod install` after creating a Podfile.

Building without pods seems to work fine, but if async-storage is included with pods, you get the error shown above.

**Downgrading to 1.3.0 fixes the issue.**

## Environment

- **Async Storage version**: 1.3.1
- **React-Native version**: 0.59.4
- **Platform tested**: iOS only

dblazeski:
Thanks @tido64 for the work on this, there's import path typo in the import statement when using Pods.

### DIFF
--- a/ios/RNCAsyncStorage.h
+++ b/ios/RNCAsyncStorage.h
@@ -7,7 +7,7 @@
 
 #import <React/RCTBridgeModule.h>
 #import <React/RCTInvalidating.h>
-#import <RNCAsyncStorage/RNCAsyncStorageDelegate.h>
+#import <RNCAsyncStorageDelegate.h>
 
 /**
  * A simple, asynchronous, persistent, key-value storage system designed as a


tido64:
Thanks for your contribution, @dblazeski. Sorry for the inconvenience this has caused.

This was also brought up by @stigi here: https://github.com/react-native-community/react-native-async-storage/pull/35#issuecomment-482604705. We discussed this a bit and agreed to rename the `.podspec` to be more in line with Apple naming guidelines. Would you be willing to make the changes since you're already here? I suppose you could also wait but that's entirely up to you.

Thanks again ❤️

Issue #76, opened 2019-04-13
Xcode archive generates Generic Xcode Archive instead of IPA


mohammadalijf:
## Current behavior

after updating to 1.3.1 Xcode creates Generic Xcode Archive instead of creating IPA. after looking up the problem and checking created archive, i found out `RNCAsyncStorage.h` and `RNCAsyncStorageDelegate.h` which are header files for the library is included in archive. you can find headers in 
Archive.xcarchive -> show Package Content/Products/usr/local/include

after looking up the problem i found a solution [here](https://stackoverflow.com/a/48408047/2161945) which is to move headers from Public to Project in the Build Phase / Headers.

## Expected behavior

Xcode should create IPA not a Generic Xcode Archive 

## Repro steps

- Xcode -> Clean
- Xcode -> clean DerivedData folder
- Xcode -> Archive

## Environment

- **Async Storage version**: 1.3.1
- **React-Native version**: 0.59.4
- **Platform tested**: iOS

tido64:
Thanks for the report! The fix was merged and released with v1.3.3. Please upgrade and let us know if you hit any other issues. 😄

PR #77, opened 2019-04-13
fix: Exclude headers from iOS archive.

tido64:
Summary:
---------

iOS headers are being included in the final archive, making Xcode generate generic archives instead of IPAs. Moving the headers to project only should address this issue. See #76.

Test Plan:
----------

In an app that manually links to AsyncStorage, archiving should generate an IPA and not a generic archive.


tido64:
@mohammadalijf Can you verify that this works for you? Thanks 😄

mtt87:
I manually tested moving the headers from `public` to `project` and I can confirm it works fine 
Thanks @tido64

tido64:
Thanks, everyone ❤️

PR #95, opened 2019-05-01
chore: upgrade detox

### LINKED ISSUE #94: Current detox version doesn't support Xcode 10.2
## Current behavior
when I run yarn install, it fails at detox post install script 

## Expected behavior

yarn install should not fail

## Repro steps

1. clone repo and run yarn

chakrihacker:
Upgrade detox:
---------

Upgrade detox for xcode 10 and above support

closes #94

### DIFF
--- a/example/android/app/build.gradle
+++ b/example/android/app/build.gradle
@@ -159,8 +159,8 @@ dependencies {
     // tests
     androidTestImplementation project(":detox")
     androidTestImplementation 'junit:junit:4.12'
-    androidTestImplementation 'com.android.support.test:runner:1.0.2'
-    androidTestImplementation 'com.android.support.test:rules:1.0.2'
+    androidTestImplementation 'androidx.test:runner:1.1.1'
+    androidTestImplementation 'androidx.test:rules:1.1.1'
 }
 
 // Run this once to be able to run the application with BUCK
--- a/example/android/app/build.gradle
+++ b/example/android/app/build.gradle
@@ -159,8 +159,6 @@ dependencies {
     // tests
     androidTestImplementation ('com.wix:detox:+') { transitive = true }
     androidTestImplementation 'junit:junit:4.12'
-    // androidTestImplementation 'androidx.test:runner:1.1.1'
-    // androidTestImplementation 'androidx.test:rules:1.1.1'
 }
 
 // Run this once to be able to run the application with BUCK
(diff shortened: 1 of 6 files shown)


chakrihacker:
Android now uses android test implementation 
https://github.com/wix/Detox/blob/master/docs/Guide.Migration.md
Followed above guide to migrate versions

example/android/app/build.gradle
```diff
@@ -117,8 +117,7 @@ android {
         versionCode 1
         versionName "1.0"
         testBuildType System.getProperty('testBuildType', 'debug')
-        missingDimensionStrategy "minReactNative", "minReactNative46"
-        testInstrumentationRunner "android.support.test.runner.AndroidJUnitRunner"
+        testInstrumentationRunner "androidx.test.runner.AndroidJUnitRunner"
```
Krizzu: Any particular reason to move to `androidx`? I haven't seen anything in Detox's migration guide that mention it. I've seen (and had) a bad experience with mixing support libs with `androidx` in one project, so ideally we'd love to avoid that conflict.

example/android/app/build.gradle
```diff
@@ -157,10 +157,10 @@ dependencies {
     implementation "com.facebook.react:react-native:+"  // From node_modules
 
     // tests
-    androidTestImplementation project(":detox")
+    androidTestImplementation ('com.wix:detox:+') { transitive = true }
     androidTestImplementation 'junit:junit:4.12'
-    androidTestImplementation 'com.android.support.test:runner:1.0.2'
-    androidTestImplementation 'com.android.support.test:rules:1.0.2'
+    // androidTestImplementation 'androidx.test:runner:1.1.1'
```
Krizzu: Comment left out.
Do we need those deps?

tido64:
@Krizzu: Regarding Xcode, one should be on at least 10.1 nowadays to build for latest iOS. But isn't this just if you need to compile the tests? Regular consumers should not be affected, correct?

example/android/app/build.gradle
chakrihacker: When you migrate from 10.x to 11.x androidx is necessary for detox

example/android/app/build.gradle
chakrihacker: I think it's not needed, let me build and check in local once to confirm this

chakrihacker:
as @tido64 mentioned, it's necessary only for contributors, it doesn't effect users

example/android/app/build.gradle
chakrihacker: removed comments

PR #98, opened 2019-05-08
[Android] Make db size configurable 

alexkuttig:
Summary:
---------

Since I do need a larger max db size, I added an optional size parameter to the package constructor. This makes the maximum db size configurable, while keeping the standard size of 6L * 1024L * 1024L when not giving any value. (like discussed in #https://github.com/react-native-community/react-native-async-storage/issues/83)

> new AsyncStoragePackage(100L * 1024L * 1024L)

Test Plan:
----------

1.) create async storage in MainApplication.java of the test app with new AsyncStoragePackage()
2.) write > 6MB data to async storage
3.) change initialization of AsyncStorage in MainApplication.java to new AsyncStoragePackage(100L * 1024L * 1024L)
4.) write additional data

Issue #83, opened 2019-04-17
Why 6mb dataset size for android?


Nthalk:
I want to know why we carry forward a non-configurable, arbitrary value (6mb) for the maximum database size for Android, while leaving IOS unlimited.

Especially when I'll just get around this with JSON stringify and react-native-sqlite-storage.

https://github.com/react-native-community/react-native-async-storage/blob/c57cbeb25034c4291cc148084ad3933266daf641/android/src/main/java/com/reactnativecommunity/asyncstorage/ReactDatabaseSupplier.java#L46

Krizzu:
Hey @Nthalk,

We did not change the initial limit, because of [a note left from original creators](https://github.com/react-native-community/react-native-async-storage/blob/c57cbeb25034c4291cc148084ad3933266daf641/android/src/main/java/com/reactnativecommunity/asyncstorage/ReactDatabaseSupplier.java#L103-L105). I'm not quite sure if it's still relevant or not, but I think we could let this value be set from user (for example, a new limit passed as argument to `AsyncStorage` package in `MainApplication`.

[FYI, here are limits of SQLite DB.](https://www.sqlite.org/limits.html)


thanks.

tido64:
I'm closing this issue since we've just merged a change to make this limit configurable. Please try it out and let us know how you go.

PR #99, opened 2019-05-10
[Android] Make db size configurable

alexkuttig:
Summary:
---------

Since I do need a larger max db size, I added an optional parameter to gradle properties. This makes the maximum db size configurable, while keeping the standard size of 6L * 1024L * 1024L when not giving any value. (like discussed in #83 and #98 )

Test Plan:
----------

1.) Use test app
2.) write > 6MB data to async storage
3.) change android/gradle.properties of testapp and uncomment the last line
4.) make new build
5.) write additional data


alexkuttig:
@Krizzu I made a new Pull Request to keep the files clean.

example/android/gradle.properties
```diff
@@ -16,3 +16,6 @@
 # This option should only be used with decoupled projects. More details, visit
 # http://www.gradle.org/docs/current/userguide/multi_project_builds.html#sec:decoupled_projects
 # org.gradle.parallel=true
+
+# If you know what you're doing, you can uncomment this line and increase the size of db
```
Krizzu: Let's be more gentle :)

```suggestion
# This is an example of how you can change default DB size (6MB) to 10MB
```

alexkuttig:
Here we go :-)

tido64:
LGTM. I'm not as savvy on Android as I'd like to be so if @Krizzu says to merge, then so be it. 👍

mjmasn:
BTW this was already possible via `onCreate()` in `MainApplication.java`:
```java
@Override
public void onCreate() {
    super.onCreate();
    long size = 25L * 1024L * 1024L;
    ReactDatabaseSupplier.getInstance(getApplicationContext()).setMaximumSize(size);
}
```

Not sure if it's worth adding that to the documentation as well?

Krizzu:
@mjmasn Hey,

I don't think it's worth adding. Setting values in `gradle.properties` are more user friendly IMO.

PR #111, opened 2019-05-25
fix compilation error

ShaMan123:
When building DexMerger throws an error due to duplicate imports of the package

Summary:
---------

Test Plan:
----------

build a release version


Krizzu:
Thanks 🙏 	

I did not get that error before but I guess we're good to prevent it by merging it.

cc @tido64

tido64:
Thanks for fixing this. 😄

PR #114, opened 2019-05-30
chore: add things to try for common issues

tido64:
Summary:
---------

Added a page with things to try out for some build related issues.

Test Plan:
----------

n/a


Krizzu:
Great 💪 	

What do you think about a section for `[@RNC/AsyncStorage]: NativeModule: AsyncStorage is null` error? Instructions for manually checking if AS is linked correctly could be included.

PR #119, opened 2019-06-07
Fixes 'RNCAsyncStorage/RNCAsyncStorageDelegate.h' file not found

harveyconnor:
Summary:
---------

Was receiving a compile error when running `react-native run-ios`
```
node_modules/@react-native-community/async-storage/ios/RNCAsyncStorage.h:10:9:
'RNCAsyncStorage/RNCAsyncStorageDelegate.h' file not found

#import <RNCAsyncStorage/RNCAsyncStorageDelegate.h>
            ~~ ~~~~~^~~
```
A similar fix was required in other libraries, since the file is in the same directory as the header. You can require it without a namespace.

Test Plan:
----------

Run `react-native run-ios` and if it compiles then it was a success.

### DIFF
--- a/ios/RNCAsyncStorage.h
+++ b/ios/RNCAsyncStorage.h
@@ -7,7 +7,7 @@
 
 #import <React/RCTBridgeModule.h>
 #import <React/RCTInvalidating.h>
-#import <RNCAsyncStorage/RNCAsyncStorageDelegate.h>
+#import "RNCAsyncStorageDelegate.h"
 
 /**
  * A simple, asynchronous, persistent, key-value storage system designed as a


Krizzu:
Thanks!

cc @tido64

tido64:
Thanks :)

Did you try this in a project with modules disabled? I think we added the full path because of it.

harveyconnor:
Nope this was with other modules present.

tido64:
Sorry, I meant Objective-C modules. You can see the setting in Xcode: https://i.stack.imgur.com/PyWgB.png

tido64:
Also, which version of AsyncStorage are you using? This shouldn't be happening in the latest version.

harveyconnor:
Modules was enabled.
And using version 1.4.2

PR #120, opened 2019-06-13
fix: Remove unnecessary warning on tvOS

bcontrol:
Summary:
---------

Since react-native-async-storage on tvOS works without problem, we should remove unnecessary warning

Test Plan:
----------

Run code on tvOS


Krizzu:
Hey @bcontrol 	
`Since react-native-async-storage on tvOS works without problem`.
Can you elaborate more?

bcontrol:
Persist storage is supported in tvOS. But yes, it's very limited. Let's leave this warning.

PR #136, opened 2019-06-22
[v2] Initial Core implementation

Krizzu:
Summary:
---------

Todo:

- [X] Initial implementation
- [x] Write tests
- [x] CI to run tests

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -137,6 +137,14 @@ jobs:
           name: Type check
           command: yarn test:types
 
+  "Test: unit":
+    <<: *js_defaults
+    steps:
+      - *addWorkspace
+      - run:
+          name: Unit tests run
+          command: yarn test:unit
+
   "Test: iOS e2e":
     <<: *macos_defaults
     steps:
@@ -289,5 +297,8 @@ workflows:
           requires:
             - "Setup environment"
       - "Test: types":
+          requires:
+            - "Setup environment"
+      - "Test: unit":
           requires:
             - "Setup environment"
--- a/babel.config.js
+++ b/babel.config.js
@@ -5,7 +5,7 @@ module.exports = {
       '@babel/preset-env',
       {
         targets: {
-          node: 'current',
+          node: 8,
         },
       },
     ],
--- a/babel.config.js
+++ b/babel.config.js
@@ -1,5 +1,15 @@
 module.exports = {
-  presets: ['@babel/preset-typescript'],
+  presets: [
+    '@babel/preset-typescript',
+    [
+      '@babel/preset-env',
+      {
+        targets: {
+          node: 'current',
+        },
+      },
+    ],
+  ],
   plugins: [
     '@babel/proposal-class-properties',
     '@babel/proposal-object-rest-spread',
--- a/core/__tests__/AsyncStorage.test.ts
+++ b/core/__tests__/AsyncStorage.test.ts
@@ -1,0 +1,121 @@
+import AsyncStorage from '../src/AsyncStorage';
+
+class StorageMock implements IStorageBackend<any> {
+  getSingle = jest.fn();
+  setSingle = jest.fn();
+  getMany = jest.fn();
+  setMany = jest.fn();
+  removeSingle = jest.fn();
+  removeMany = jest.fn();
+  getKeys = jest.fn();
+  dropStorage = jest.fn();
+}
+
+describe('AsyncStorage', () => {
+  const mockedStorage = new StorageMock();
+
+  beforeEach(() => {
+    jest.resetAllMocks();
+  });
+
+  describe('main API', () => {
+    it('handles basic set/read/remove calls', async () => {
+      const as = new AsyncStorage(mockedStorage, {
+        logger: false,
+        errorHandler: false,
+      });
+
+      const key = 'myKey';
+      const value = {
+        name: 'Jerry',
+      };
+
+      // get
+      await as.get(key);
+      expect(mockedStorage.getSingle).toBeCalledWith(key, null);
+
+      //set
+      await as.set(key, value);
+      expect(mockedStorage.setSingle).toBeCalledWith(key, value, null);
+
+      // remove
+      await as.remove(key);
+      expect(mockedStorage.removeSingle).toBeCalledWith(key, null);
+    });
+
+    it('handles basic multi set/read/remove calls', async () => {
+      const keys = ['key1', 'key2', 'key3'];
+      const keyValues = [{key1: 'value1'}, {key2: 'value2'}, {key3: 'value3'}];
+
+      const as = new AsyncStorage(mockedStorage, {
+        logger: false,
+        errorHandler: false,
+      });
+
+      // set
+      await as.setMultiple(keyValues);
+      expect(mockedStorage.setMany).toBeCalledWith(keyValues, null);
+
+      // get
+      await as.getMultiple(keys);
+      expect(mockedStorage.getMany).toBeCalledWith(keys, null);
+
+      // remove
+      await as.removeMultiple(keys);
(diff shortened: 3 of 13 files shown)


.circleci/config.yml
```diff
@@ -137,6 +137,14 @@ jobs:
           name: Type check
           command: yarn test:types
 
+  "Test: unit":
```
thymikee: FYI future notice: use the community orb for easier circle management: https://github.com/react-native-community/react-native-circleci-orb

babel.config.js
```diff
@@ -1,5 +1,15 @@
 module.exports = {
-  presets: ['@babel/preset-typescript'],
+  presets: [
+    '@babel/preset-typescript',
+    [
+      '@babel/preset-env',
+      {
+        targets: {
+          node: 'current',
```
thymikee:
```suggestion
          node: 8.3,
```
RN requires Node 8.3, we should be compatible

core/__tests__/AsyncStorage.test.ts
```diff
@@ -0,0 +1,121 @@
...
+
+describe('AsyncStorage', () => {
+  const mockedStorage = new StorageMock();
+
+  beforeEach(() => {
+    jest.resetAllMocks();
+  });
+
+  describe('main API', () => {
+    it('handles basic set/read/remove calls', async () => {
```
thymikee: I'd use [test.each](https://jestjs.io/docs/en/api#testeachtable-name-fn-timeout) for these tests for convenience

core/__tests__/AsyncStorage.test.ts
```diff
@@ -0,0 +1,121 @@
...
+        logger: false,
+        errorHandler: false,
+      });
+
+      const key = 'myKey';
+      const value = {
+        name: 'Jerry',
+      };
+
+      // get
```
thymikee: these comments are useless, please remove them all

core/__tests__/AsyncStorage.test.ts
```diff
@@ -0,0 +1,121 @@
...
+    });
+
+    it('handles getKeys/clearStorage/instance calls', async () => {
+      const asyncStorage = new AsyncStorage(mockedStorage, {
+        logger: false,
+        errorHandler: false,
+      });
+
+      // instance
+      expect(asyncStorage.instance()).toBe(mockedStorage);
```
thymikee: is this a part of your public API? `asyncStorage` is already an instance of a class, so this is quite unintuitive

core/__tests__/AsyncStorage.test.ts
```diff
@@ -0,0 +1,121 @@
...
+      it('uses logger when provided', async () => {
+        const loggerFunc = jest.fn();
+
+        const as = new AsyncStorage(mockedStorage, {
+          logger: loggerFunc,
+          errorHandler: false,
+        });
+
+        await as.get('key');
+        expect(loggerFunc).toBeCalledTimes(1);
```
thymikee: called with what?

core/__tests__/AsyncStorage.test.ts
Krizzu: oops 🤦‍♂

core/__tests__/AsyncStorage.test.ts
```diff
@@ -0,0 +1,121 @@
...
+
+        const as = new AsyncStorage(mockedStorage, {
+          errorHandler,
+          logger: false,
+        });
+
+        await as.get('key');
+
+        expect(errorHandler).toBeCalledTimes(1);
+        expect(errorHandler.mock.calls[0][0]).toBe(error);
```
thymikee:
```suggestion
        expect(errorHandler).toHaveBeenCalledWith(error);
```

core/__tests__/AsyncStorage.test.ts
Krizzu: Yes, it is. Some storages might come with other functionalities than `AsyncStorage` requires, so `instance` gives direct access to the storage (and to said functionalities)

core/__tests__/core.test.ts
```diff
@@ -0,0 +1,84 @@
+import Factory from '../src/';
+import {simpleLogger, simpleErrorHandler} from '../src/defaults';
+
+describe('AsyncStorageFactory', () => {
+  it('Throws when tried to instantiate', () => {
+    expect(() => new Factory()).toThrow();
```
thymikee: please add some text that you expect to throw, it may have thrown for other reasons and test will pass, something like:
```suggestion
    expect(() => new Factory()).toThrow('forbidden to instantiate Factory');
```

core/__tests__/AsyncStorage.test.ts
thymikee: Hm, this is dangerous. All functionalities should be behind a facade of your API, don't you think?

Krizzu: That'd be ideal, but we cannot predict all possibilities, right?

thymikee: I'd be cautious about that, this is what Enzyme does that sooner or later made everyone using this API because guides and blog post were made. I'd remove this and think about expanding API when necessary

core/__tests__/core.test.ts
```diff
@@ -0,0 +1,84 @@
...
+  afterAll(() => {
+    console.error.mockRestore();
+  });
+  it('logs error when it is a string', () => {
+    const errorMessage = 'Fatal!';
+
+    simpleErrorHandler(errorMessage);
+
+    expect(console.error).toBeCalledTimes(1);
+    expect(console.error.mock.calls[0][0]).toEqual(errorMessage);
```
thymikee:
```suggestion
    expect(console.error).toHaveBeenCalledWith(errorMessage);
```

core/types/index.d.ts
```diff
@@ -1,9 +1,110 @@
-interface IStorageBackend {
-  getSingle(): Promise<any>;
-  //todo
+/**
+ *
+ * Core Async Storage API
+ *
+ */
+declare class AsyncStorage<
```
thymikee: can't these types be output by tsc?

jest.config.js
```diff
@@ -0,0 +1,19 @@
+const commonSettings = {
+  transform: {
+    '^.+\\.tsx?$': 'babel-jest',
```
thymikee: this is the default transform, you should be good to remove it

core/__tests__/AsyncStorage.test.ts
Krizzu: Going to leave it like that for now, with a note to re-think it later.

core/types/index.d.ts
Krizzu: They were at first, but then I decided to remove it, as I needed more control over it

Krizzu:
Thanks a lot!

core/__tests__/AsyncStorage.test.ts
```diff
@@ -0,0 +1,160 @@
...
+          errorHandler: false,
+        });
+
+        const key = 'myKey';
+        const value = {
+          name: 'Jerry',
+        };
+
+        switch (methodName) {
+          case 'get': {
```
thymikee: Yikes, that's not what I meant, this code is almost no different. You can use each like this:
```js
it.each([
  ['get', 'myKey'], 
  ['set', 'myKey', {name: 'Jerry'}], 
  ['remove', 'myKey']
])('handles single %s api call', async (methodName: 'get' | 'set' | 'remove', ...args) => {
  await as[methodName](...args);
  expect(mockedStorage.getSingle).toBeCalledWith(...args.concat(null));
}
```

Krizzu: yh, shame. Will cover this in next PR.

PR #146, opened 2019-07-02
Add trailing `/` to `transformIgnorePatterns`

genuinezack:
Fix a small documentation issue with the `transformIgnorePatterns` suggested fix for performing proper transformations on the mocks for Jest compatibility. Need to add a trailing `/` to the regex for the transform ignore, otherwise it will match all packages in `node-modules` that start with the string `react-native`!

### DIFF
--- a/docs/Jest-integration.md
+++ b/docs/Jest-integration.md
@@ -71,17 +71,18 @@ You can [check its implementation](../jest/async-storage-mock.js) to get more in
 
 You need to point Jest to transform this package. You can do so, by adding Async Storage path to `transformIgnorePatterns` setting in Jest's configuration.
 
-
 ```json
 "jest": {
   "transformIgnorePatterns": ["node_modules/(?!(@react-native-community/async-storage/lib))"]
 }
 ```
 
-Optionally, you can transform whole scope for `react-native-community` and `react-native`:
+Optionally, you can transform all packages in the `react-native-community` and `react-native` directories by specifying this pattern:
 
 ```json
 "jest": {
   "transformIgnorePatterns": ["node_modules/(?!(@react-native-community|react-native)/)"]
 }
 ```
+
+Or you can expand it even further to transform all packages whose names begin with `react-native` by omitting the trailing `/` in the above pattern.
--- a/docs/Jest-integration.md
+++ b/docs/Jest-integration.md
@@ -82,6 +82,6 @@ Optionally, you can transform whole scope for `react-native-community` and `reac
 
 ```json
 "jest": {
-  "transformIgnorePatterns": ["node_modules/(?!(@react-native-community|react-native))"]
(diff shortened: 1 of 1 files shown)


Krizzu:
Hey @genuinezack , thanks for your contribution!

This looks legit to me, although, I'll ask for a second opinion :) cc @thymikee

thymikee:
@genuinezack this behavior is expected, as the vast majority of `react-native-*` packages are not compiled to JSC-compatible code, but only expose sources. It's a shame, but this is the reality we live in currently. I don't think we should merge this.

genuinezack:
I disagree that this is expected/desirable, @thymikee, for three reasons:
1) The line before this snippet in the documentation says this line will transform "the whole scope of `react-native-community` and `react-native`". I understand (or misunderstand) that to mean those scopes **specifically** and not all scopes that happen to start with those strings.
2) Casting a wide net like this is likely to have false positives and negatives, meaning developers will have to edit this line anyways. They'll have an easier time reasoning about which packages are included if this line matches specific packages not string prefixes.
3) Finally, [Jest's recommended pattern](https://jestjs.io/docs/en/tutorial-react-native#transformignorepatterns-customization) for this line does include a trailing slash for packages. Since it's their tool, I'd argue we should agree with them.

That being said, it is a single character in a legacy documentation so, as I noted, no concern to my life if you leave it as-is, I'll still use the package and appreciate you for making and maintaining it.

thymikee:
>Casting a wide net like this is likely to have false positives and negatives, meaning developers will have to edit this line anyways. 

Never happened to me in 2 years

>They'll have an easier time reasoning about which packages are included if this line matches specific packages not string prefixes.

`transformIgnorePatterns` is one of the worst APIs Jest provides and believe me it causes constant confusion, especially for people starting with React Native. 

Fun fact – Metro (the React Native bundler) transpiles _all_ of your dependencies already with the same Babel preset.

I get your point though, it's completely correct. I'm just too tired answering the same questions (I'm a Jest maintainer btw). What would be a nice compromise I guess, is leaving this regex as you like, explain what it does, and how to make it more permissive. It's not quite the scope of this package, but maybe it helps someone :)

genuinezack:
@thymikee - I'm only 9 months into the world of react-native, but we ran into this recently with `cavy` and `@segment/analytics-react-native`, both of which needed to be added to this and is why this particular regex is top of mind for me 😅 But, relatedly, 100% agree with your comment that it is a source of constant confusion!

I updated my change with your suggested clarification and will leave the compromise in your court to approve/deny.

(Oh also thanks for maintaining Jest too! I have no idea how y'all open source maintainers deal with hearing the same issues/suggestions/requests all day long, so sorry to add to the din.)

Krizzu:
Right, this should resolve any doubts about this regexp :) I'm going to look into android failing test later today and merge it.

Krizzu:
Can't figure out why Android e2e fails at deps installations, so, for now, I'll go ahead and just merge this.

PR #160, opened 2019-07-10
The AsyncTask.THREAD_POOL_EXECUTOR maybe blocked by some other module

gaodeng:
fix  https://github.com/react-native-community/async-storage/issues/159

### DIFF
--- a/android/build.gradle
+++ b/android/build.gradle
@@ -30,6 +30,10 @@ if( newDbSize != null && newDbSize.isLong()) {
     dbSizeInMB = newDbSize.toLong()
 }
 
+def useAsyncTaskThreadPoolExecutor = rootProject.hasProperty('AsyncStorage_use_AsyncTaskThreadPoolExecutor') ? rootProject.properties['AsyncStorage_use_AsyncTaskThreadPoolExecutor'] : true
+
+
+
 apply plugin: 'com.android.library'
 
 android {
@@ -41,6 +45,7 @@ android {
         targetSdkVersion getExtOrIntegerDefault('targetSdkVersion')
 
         buildConfigField "Long", "AsyncStorage_db_size", "${dbSizeInMB}L"
+        buildConfigField("boolean", "AsyncStorage_use_AsyncTaskThreadPoolExecutor", "${useAsyncTaskThreadPoolExecutor}")
     }
 }
 
--- a/android/build.gradle
+++ b/android/build.gradle
@@ -30,9 +30,9 @@ if( newDbSize != null && newDbSize.isLong()) {
     dbSizeInMB = newDbSize.toLong()
 }
 
-def useAsyncTaskThreadPoolExecutor = rootProject.hasProperty('AsyncStorage_use_AsyncTaskThreadPoolExecutor') ? rootProject.properties['AsyncStorage_use_AsyncTaskThreadPoolExecutor'] : true
-
-
+def useCustomExecutor = rootProject.hasProperty('AsyncStorage_useCustomExecutor')
+        ? rootProject.properties['AsyncStorage_useCustomExecutor']
+        : false
 
 apply plugin: 'com.android.library'
 
@@ -45,7 +45,7 @@ android {
         targetSdkVersion getExtOrIntegerDefault('targetSdkVersion')
 
         buildConfigField "Long", "AsyncStorage_db_size", "${dbSizeInMB}L"
-        buildConfigField("boolean", "AsyncStorage_use_AsyncTaskThreadPoolExecutor", "${useAsyncTaskThreadPoolExecutor}")
+        buildConfigField("boolean", "AsyncStorage_useCustomExecutor", "${useCustomExecutor}")
     }
 }
 
--- a/android/src/main/java/com/reactnativecommunity/asyncstorage/AsyncStorageModule.java
+++ b/android/src/main/java/com/reactnativecommunity/asyncstorage/AsyncStorageModule.java
@@ -28,13 +28,8 @@ import com.facebook.react.modules.common.ModuleDataCleaner;
 
 import java.util.ArrayDeque;
 import java.util.HashSet;
-import java.util.concurrent.BlockingQueue;
 import java.util.concurrent.Executor;
-import java.util.concurrent.LinkedBlockingQueue;
-import java.util.concurrent.ThreadFactory;
-import java.util.concurrent.ThreadPoolExecutor;
-import java.util.concurrent.TimeUnit;
-import java.util.concurrent.atomic.AtomicInteger;
+import java.util.concurrent.Executors;
 
 @ReactModule(name = AsyncStorageModule.NAME)
 public final class AsyncStorageModule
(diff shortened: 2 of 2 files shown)


Krizzu:
@gaodeng Thanks for your contribution. Does it fix the issue on your devices?

Extending the number of threads available would make this feature more resource-hungry. I'd say to hide this implementation behind the flag, making it opt-in feature.

gaodeng:
I am not an android expert, here are my immature ideas.

I think we should not use the shared AsyncTask.THREAD_POOL_EXECUTOR .

Here are the reasons

- AsyncStorage is a simple key-value storage system. It should be very fast and reliable. AsyncTask.THREAD_POOL_EXECUTOR may be occupied by other modules, causing AsyncStorage operations to wait for tasks from other modules to complete. This makes Async Storage's operating speed unreliable. This waiting time can be very long. For example, we used the react native fbsdk module. When users in China use the app, Async Storage access often becomes very slow. It was later discovered that because the react native fbsdk used AsyncTask.THREAD_POOL_EXECUTOR, because Facebook could not be accessed in China, these tasks all had to wait for the network timeout, which caused AsyncTask.THREAD_POOL_EXECUTOR to be blocked.

- Using a separate THREAD_POOL_EXECUTOR does not have much overhead, but it will make AsyncStorage more reliable.

Sorry for any mistakes. English is not my native language

Krizzu:
@gaodeng Right, thanks for the explanation. 

These are all valid points and I think it makes sense to have your changes merged. My only requirement for this to land is to hide this feature behind a flag, [like we did for size limit for android](https://github.com/react-native-community/async-storage/blob/LEGACY/docs/advanced/IncreaseDbSize.md), so devs can opt-out of standard thread pool provided from Async Storage. The flag would [be read in this place](https://github.com/react-native-community/async-storage/pull/160/files#diff-82d7ea36ea8e23881bed5082f1974af4R112), deciding which thread pool to use.

Let me know if you'd need some help implementing it.

android/src/main/java/com/reactnativecommunity/asyncstorage/AsyncStorageModule.java
```diff
@@ -44,6 +50,31 @@
...
+    public Thread newThread(Runnable r) {
+      return new Thread(r, "AsyncStorageTask #" + mCount.getAndIncrement());
+    }
+  };
+
+  private static final BlockingQueue<Runnable> sPoolWorkQueue =
+          new LinkedBlockingQueue<Runnable>(128);
+
+  static {
+    ThreadPoolExecutor threadPoolExecutor = new ThreadPoolExecutor(
```
tido64: Should we be using one of the `Executors` factory methods as mentioned in [the documentation](https://developer.android.com/reference/java/util/concurrent/ThreadPoolExecutor) instead? The  [`Executors#newSingleThreadExecutor`](https://developer.android.com/reference/java/util/concurrent/Executors.html#newSingleThreadExecutor()) sounds like a good fit to me as it will effectively serialize all storage operations.

tido64: @Krizzu, I think using one of the factory methods will vastly simplify this PR.

Krizzu: Yup, I agree - sequential operations make more sense to be used here. 

@gaodeng Can you amend your changes?

gaodeng: sure,thanks for the tip

yazhengwang: @gaodeng 谢谢大兄弟，终于解决了这个问题

android/src/main/java/com/reactnativecommunity/asyncstorage/AsyncStorageModule.java
```diff
@@ -78,7 +80,7 @@ synchronized void scheduleNext() {
   private final SerialExecutor executor;
 
   public AsyncStorageModule(ReactApplicationContext reactContext) {
-    this(reactContext, AsyncTask.THREAD_POOL_EXECUTOR);
+    this(reactContext, BuildConfig.AsyncStorage_use_AsyncTaskThreadPoolExecutor?AsyncTask.THREAD_POOL_EXECUTOR:Executors.newSingleThreadExecutor());
```
tido64: Nit: Reformat for better legibility.
```suggestion
    this(
      reactContext,
      BuildConfig.AsyncStorage_use_AsyncTaskThreadPoolExecutor
        ? AsyncTask.THREAD_POOL_EXECUTOR
        : Executors.newSingleThreadExecutor());
```

android/src/main/java/com/reactnativecommunity/asyncstorage/AsyncStorageModule.java
```diff
@@ -44,6 +45,7 @@
   private ReactDatabaseSupplier mReactDatabaseSupplier;
   private boolean mShuttingDown = false;
 
+
```
tido64: Nit: Remove superfluous newline.
```suggestion
```

android/build.gradle
```diff
@@ -30,6 +30,10 @@ if( newDbSize != null && newDbSize.isLong()) {
     dbSizeInMB = newDbSize.toLong()
 }
 
+def useAsyncTaskThreadPoolExecutor = rootProject.hasProperty('AsyncStorage_use_AsyncTaskThreadPoolExecutor') ? rootProject.properties['AsyncStorage_use_AsyncTaskThreadPoolExecutor'] : true
```
tido64: Nit: I'm not really fond of explicitly returning `true`/`false` in if-else statements.
```suggestion
def useAsyncTaskThreadPoolExecutor = !rootProject.hasProperty('AsyncStorage_use_AsyncTaskThreadPoolExecutor') || rootProject.properties['AsyncStorage_use_AsyncTaskThreadPoolExecutor']
```

android/build.gradle
```diff
@@ -30,6 +30,10 @@ if( newDbSize != null && newDbSize.isLong()) {
...
+
+
+
```
tido64: Can we remove these superfluous newlines?

android/build.gradle
tido64: Should we consider having this disabled by default? @Krizzu?

Krizzu: Yes, disabled by default please

android/build.gradle
```diff
@@ -41,6 +45,7 @@ android {
         targetSdkVersion getExtOrIntegerDefault('targetSdkVersion')
 
         buildConfigField "Long", "AsyncStorage_db_size", "${dbSizeInMB}L"
+        buildConfigField("boolean", "AsyncStorage_use_AsyncTaskThreadPoolExecutor", "${useAsyncTaskThreadPoolExecutor}")
```
Krizzu: nit: could we rename it to something less typo-prone name? Like `AsyncStorage_customExecutor` ?

android/build.gradle
gaodeng: disable customExecutor or AsyncTaskThreadPoolExecutor by default ?

Krizzu: Good point, it's confusing now. To clear it up:

1. Change the name of the property to `AsyncStorage_useCustomExecutor`
2. The default thread pool executor (without setting the property to `true`) should be the one provided from `AsyncTask`, so please change it to:

```
def useCustomExecutor = rootProject.hasProperty('AsyncStorage_useCustomExecutor') 
? rootProject.properties['AsyncStorage_useCustomExecutor'] 
: false
```

3. Use `useCustomExecutor` variable to decide about Executor - `true` to use `SingleThreadedExecutor`, false for `AsyncTask`'s executor

android/build.gradle
Krizzu: @gaodeng Please just remove this extra line 🙏

gaodeng: @Krizzu  line 32 or line 36 ?
Screen Shot 2019-07-21 at 7 59 07 PM

yazhengwang:
@gaodeng @Krizzu   我发现在Android 6.0.1,level 23 还是有问题，读取还是会等待很长时间，但是在 Android9.0 上是 OK 的
image

kangbb:
I'm the Expo user, so I must use the `AsyncStorage` from `react-native`. And it's also have the problem on some Android device. Is there any way to fix it?

a396901990:
same as me first-time use is slow on some android devices (Xiaomi 8, Samsang S6 Edge), feel like the issue not solved

a396901990:
solved with https://github.com/react-native-community/async-storage/blob/LEGACY/docs/advanced/DedicatedExecutor.md

kangbb:
> solved with https://github.com/react-native-community/async-storage/blob/LEGACY/docs/advanced/DedicatedExecutor.md

Thanks for your answer. But I'm a Expo user, emm...., which means I can't use any nativemodule, this workaround not useful. Maybe I should exit Expo.

tido64:
@kangbb If you want to have more control over your native modules, you [have to eject](https://docs.expo.io/versions/latest/expokit/eject/#should-i-eject-to-expokit). The alternative is to have Expo embed community AsyncStorage but I don't know whether it's on their roadmap.

PR #161, opened 2019-07-14
[v2] Setup Example app

Krizzu:
Summary:
---------

- [X] Working examples (both platforms)
- [x] Better example app
- [x] Clean up


Krizzu:
This should cover the fundamentals for future example apps.

PR #166, opened 2019-07-18
fix: getAllKeys() Jest mock returns array of keys

matthamil:
Summary:
---------

The `AsyncStorage.getAllKeys()` method in the Jest mock wasn't mocked to return the mock storage keys. This changes the mock such that `getAllKeys` will by default return the mock storage keys. Without this change, lib users will have to mock `getAllKeys`.

Test Plan:
----------

I don't have a perfect test plan, but you can test by writing a Jest test that will have code that executes `AsyncStorage.getAllKeys()`. The default mock will not return the mock storage keys.

### DIFF
--- a/jest/async-storage-mock.js
+++ b/jest/async-storage-mock.js
@@ -40,7 +40,7 @@ const asMock = {
   ),
 
   clear: jest.fn<[CallbackType], Promise<*>>(_clear),
-  getAllKeys: jest.fn<[], void>(),
+  getAllKeys: jest.fn<[], Promise<string[]>>(_getAllKeys),
   flushGetRequests: jest.fn<[], void>(),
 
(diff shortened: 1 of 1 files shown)


jest/async-storage-mock.js
```diff
@@ -40,7 +40,7 @@ const asMock = {
   ),
 
   clear: jest.fn<[CallbackType], Promise<*>>(_clear),
-  getAllKeys: jest.fn<[], void>(),
+  getAllKeys: jest.fn<[], Promise<string[]>>(_getAllKeys),
```
matthamil: I am unsure if the types are correct on L43.

Krizzu: Looks fine

matthamil: Thanks!

matthamil:
Thanks for looking over the PR, and thank you for the time you spend on this project :)

PR #167, opened 2019-07-21
Add tvOS to podspec supported platforms

adamkoch:
Summary:
---------

After upgrading my tvOS project to RN 0.60+, I found the new pod auto-linking feature wasn't working for async-storage on tvOS due to support not being specified for tvOS. Adding this to the podspec allows `pod install` to work correctly for tvOS.

Test Plan:
----------

Try and use async-storage in a tvOS project on RN 0.60+. Note that `pod install` will throw an error (tvOS not supported). Update the podspec as per this PR and try again and it will work.

### DIFF
--- a/RNCAsyncStorage.podspec
+++ b/RNCAsyncStorage.podspec
@@ -10,7 +10,7 @@ Pod::Spec.new do |s|
 
   s.authors      = package['author']
   s.homepage     = package['homepage']
-  s.platform     = :ios, "9.0"
+  s.platforms    = { :ios => "9.0", :tvos => "9.2" }
 
   s.source       = { :git => "https://github.com/react-native-community/react-native-async-storage.git", :tag => "v#{s.version}" }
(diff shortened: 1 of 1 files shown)


Krizzu:
Right, this looks like something we need for people using `tvOS` with autolinking. One question: Why you use version `9.2` instead of `9.0`, like `iOS` does?

Cc @tido64

adamkoch:
This is the version I tested/verified personally and that my environment was set up with. I believe when creating a new RN project the tvOS component that gets created has a target of tvOS 9.2, I'm not sure the reason why.

See:
https://github.com/facebook/react-native/blob/master/template/ios/HelloWorld.xcodeproj/project.pbxproj#L549

And original commit:
https://github.com/facebook/react-native/commit/d625a72f633ae883b78aa4d69570e9d7c05f0405

Krizzu:
Right, thanks @adamkoch, looks good to me.

PR #207, opened 2019-09-13
remove android gradle plugin

dulmandakh:
Summary:
---------

RN modules don't need to include Android Gradle Plugin because it's included in App project, and version its mismatch between app and module causes issues.

This PR removes Android Gradle Plugin from **android** or module project, and created a root project, creating build.gradle.kts file, so that we can develop with Android Studio. Also removed buildToolsVersion because it's no longer required by Android Gradle Plugin. Then changed minSdkVersion to be 16 to comply with RN.

Once this PR merged, developers need to open async-storage in Android Studio instead of android folder inside it.

Test Plan:
----------

The module will work as usual and expected


dulmandakh:
@Krizzu please review and merge

android/build.gradle
```diff
@@ -37,19 +21,19 @@ def useDedicatedExecutor = rootProject.hasProperty('AsyncStorage_dedicatedExecut
...
-        targetSdkVersion getExtOrIntegerDefault('targetSdkVersion')
+        minSdkVersion getExtOrDefault('minSdkVersion', 16)
+        targetSdkVersion getExtOrDefault('targetSdkVersion', 28)
 
         buildConfigField "Long", "AsyncStorage_db_size", "${dbSizeInMB}L"
         buildConfigField("boolean", "AsyncStorage_useDedicatedExecutor", "${useDedicatedExecutor}")
     }
 }
 
 repositories {
```
Krizzu: We don't have any `dependencies` in `android` block - do we need this?

build.gradle.kts
```diff
@@ -0,0 +1,10 @@
+buildscript {
+    repositories {
+        google()
+	jcenter()
```
Krizzu: Spacing

android/build.gradle
dulmandakh: yes, we need this, at least for test builds.

dulmandakh:
@Krizzu fixed, please merge

friederbluemle:
Hi @dulmandakh - What kind of issues were you seeing when there is a mismatch of the the app project and a module? Can you please link me to an example?

Krizzu:
@dulmandakh Actually, I think we don't need to have gradle project in the root. When developing AsyncStorage, you'd normally do it by using Android App - which we have, in `example` directory and it's already using `AsyncStorage` module as subproject.  

Can you checkout changes with `build.gradle.kts` and executables please?

dulmandakh:
@Krizzu removed Gradle wrapper

friederbluemle:
@Krizzu @dulmandakh - Please check a slightly different approach here: #209 - It achieves the same result (avoiding unnecessary dependency downloads/conflicts in an app project), but keeps the buildscript dependency in place. With this it is still possible to open the `android` folder in Android Studio, and build this library stand-alone.

Krizzu:
Closing in favor of #209

PR #209, opened 2019-09-18
Update build.gradle setup

friederbluemle:
# Summary

A slightly different approach to #207 

This wraps the Android Gradle plugin dependency in the `buildscripts` section of `android/build.gradle` in a conditional:

```
if (project == rootProject) {
    // ... (dependency here)
}
```

The Android Gradle plugin is _only_ required when opening the project _stand-alone_, not when it is _included as a dependency_. By doing this, the project opens correctly in Android Studio, and it can also be consumed as a native module dependency from an application project **without** affecting the app project (avoiding unnecessary downloads/conflicts/etc).

# Test Plan

## What's required for testing (prerequisites)?

Android Studio and/or Gradle installed locally.

## What are the steps to reproduce (after prerequisites)?

Open `android` in Android Studio:

```sh
studio android/
```

Or:

Generate a Gradle wrapper on the fly, and build on command line:

```sh
cd android/
gradle wrapper --gradle-version 5.4.1 --distribution-type all
./gradlew build
```

Successfully verified and tested on:
- Android Device (Xiaomi Mi A2)
- Android Emulator (Pixel 3a)
- iOS Device (iPhone 8)
- iOS Simulator (iPhone X)

<sub><sup>Once approved, please use a *normal* GitHub merge (i.e. **NO rebase/squash** merge) to integrate the commit(s) from the PR head branch. The changes are broken up into meaningful, [atomic commits](https://www.freshconsulting.com/atomic-commits/), and my branch should already be up-to-date with the latest base branch. If it isn't, or if you want me to change anything, please let me know, and I will update the branch as soon as possible. Thank you!</sup></sub>

### DIFF
--- a/android/build.gradle
+++ b/android/build.gradle
@@ -1,21 +1,20 @@
+def safeExtGet(prop, fallback) {
+    rootProject.ext.has(prop) ? rootProject.ext.get(prop) : fallback
+}
 
 buildscript {
-    repositories {
-        google()
-        jcenter()
+    // The Android Gradle plugin is only required when opening the android folder stand-alone.
+    // This avoids unnecessary downloads and potential conflicts when the library is included as a
+    // module dependency in an application project.
+    if (project == rootProject) {
+        repositories {
+            google()
+            jcenter()
+        }
+        dependencies {
+            classpath 'com.android.tools.build:gradle:3.5.0'
+        }
     }
-
-    dependencies {
-        classpath 'com.android.tools.build:gradle:3.4.1'
-    }
-}
-
-def getExtOrDefault(name) {
-  return rootProject.ext.has(name) ? rootProject.ext.get(name) : project.properties['RNAsyncStorage_' + name]
-}
-
-def getExtOrIntegerDefault(name) {
-  return rootProject.ext.has(name) ? rootProject.ext.get(name) : (project.properties['RNAsyncStorage_' + name]).toInteger()
 }
 
 // AsyncStorage has default size of 6MB.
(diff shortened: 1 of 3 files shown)


android/build.gradle
```diff
@@ -1,12 +1,19 @@
+def safeExtGet(prop, fallback) {
+    rootProject.ext.has(prop) ? rootProject.ext.get(prop) : fallback
+}
 
 buildscript {
```
dulmandakh: can we move this to if check?

friederbluemle: Sorry what do you mean?

friederbluemle: This is the same exact check that's used in **dozens** of projects (I just copy+pasted it) :)

dulmandakh: I would like to move repositories block inside the if check

friederbluemle: Ooh, I understand now. Okay sure, let me try.

friederbluemle: @dulmandakh - I updated it. I agree, it looks a bit cleaner. Thanks!

android/build.gradle
```diff
@@ -37,22 +44,34 @@ def useDedicatedExecutor = rootProject.hasProperty('AsyncStorage_dedicatedExecut
...
 }
 
 repositories {
-    mavenCentral()
+    mavenLocal()
+    maven {
+        // All of React Native (JS, Obj-C sources, Android binaries) is installed from npm
+        url "$rootDir/../node_modules/react-native/android"
+    }
+    maven {
```
Krizzu: We don't need this

android/build.gradle
```diff
@@ -37,22 +44,34 @@ def useDedicatedExecutor = rootProject.hasProperty('AsyncStorage_dedicatedExecut
...
-    compileSdkVersion getExtOrIntegerDefault('compileSdkVersion')
-    buildToolsVersion getExtOrDefault('buildToolsVersion')
-
+    compileSdkVersion safeExtGet('compileSdkVersion', 28)
+    buildToolsVersion safeExtGet('buildToolsVersion', '28.0.3')
     defaultConfig {
-        minSdkVersion getExtOrIntegerDefault('minSdkVersion')
-        targetSdkVersion getExtOrIntegerDefault('targetSdkVersion')
-
+        minSdkVersion safeExtGet('minSdkVersion', 19)
```
Krizzu: I'm happy keeping 19 here, but as @dulmandakh said, it would be better to comply with RN's 16.

friederbluemle: Sure, I agree it would be best to comply with 16 from react-native. Is it okay to make this change in a separate PR? Lowering the minimum SDK _could_ require other code changes. Since this PR is not about change the SDK level, it would keep these changes more focussed (and easier to review).

android/build.gradle
friederbluemle: Sure. I removed it.

friederbluemle:
@Krizzu - Thanks for the review. I addressed your comments, and also removed the two "getExtOrDefault" (I thought I removed them earlier, but missed it ;))
Regarding opening it stand-alone: Sometimes it can be useful to validate and compile the project in isolation, without any other projects or apps involved. It would also be possible to add unit tests that specifically target library code (and of course don't need an app project).

android/build.gradle
```diff
@@ -1,23 +1,22 @@
...
+    // The Android Gradle plugin is only required when opening the android folder stand-alone.
+    // This avoids unnecessary downloads and potential conflicts when the library is included as a
+    // module dependency in an application project.
+    if (project == rootProject) {
+        repositories {
+            google()
+            jcenter()
+        }
+        dependencies {
+            classpath 'com.android.tools.build:gradle:3.4.1'
```
friederbluemle: @Krizzu - Quick question: I kept the version here the same as before, but could also update it to the latest stable version (`3.5.0`) at the same time. It would have no impact on any consumer app project. OK to make the change? Or do you prefer it to be in a separate PR?

Krizzu: Yes, please upgrade to 3.5.0 in this PR

android/build.gradle
Krizzu: You're right, let's keep 19 for now

android/build.gradle
friederbluemle: Done.

Krizzu:
Thanks a lot @friederbluemle 	@dulmandakh

Issue #52, opened 2019-03-30
react-native-web implementation of AsyncStorage


cohawk:
Should the react-native-web implementation of AsyncStorage also reside in this community library?

When changing to @react-native-community/async-storage the web version of my React Native app fails on `multiMerge`

I believe it would be a simple copy paste of the localStorage code here: https://github.com/necolas/react-native-web/blob/master/packages/react-native-web/src/exports/AsyncStorage/index.js

And it appears the react-native-web maintainers agree with the functionality residing in this library - https://github.com/necolas/react-native-web/issues/1260

Thank you.

Krizzu:
Thanks @cohawk for pointing this out!

I'm moving [discussion over there](https://github.com/necolas/react-native-web/issues/1260), if you think it'd be nice to continue here, let me know.


thanks.

necolas:
Please continue the discussion here and lift any code from RNW that is necessary.

necolas:
@krizzu Please reopen this issue

necolas:
This is the last commit in React Native for Web that will include `AsyncStorage` - https://github.com/necolas/react-native-web/tree/33dc3cb01869cde556387e1cbc4e43323015b5b7/packages/react-native-web/src/exports/AsyncStorage

Krizzu:
Reopening the issue to continue to talk about making Web support first-class citizen in this repo. Currently working on implementing core of AS v2, and soon going to start on adding Storages.

brunolemos:
I see the coding work on v2 has started but apparently became stale. So how should one proceed to add web support to this repo? Master branch? Legacy?

Krizzu:
@brunolemos Once #186 is done, I think it'd be a good start to add web support.

JesseRWeigel:
@Krizzu if you would like help with adding web support, please let me know. I use RNW with AsyncStorage and am holding off on upgrading.

Krizzu:
I know it took some time but - I'm open for help with Web storage. If you're interested, please respond to this issue.

I've prepared a basic scaffold that work can be best on. Checkout to `master` branch and see `packages/storage-web` package.

Some todo list:
- Follow [instructions to create a storage backend](https://github.com/react-native-community/async-storage/blob/master/packages/core/docs/Writing_Storage_Backend.md)
- Make it compatible with both `localStorage` and `sessionStorage` (add an option to select one to use, when creating an instance of this storage)
- Add unit tests
- Update Readme

[...]

PR #216, opened 2019-09-24
[v2] Web Storage

JesseRWeigel:
Summary:
---------
#52 

I added a web-storage option that uses localStorage(default) or sessionStorage(by passing true as the first param when creating a Webstorage instance).

I also added a react app as an example to demonstrate using Webstorage. I followed the same UI structure as the React Native example app.

Test Plan:
----------

You can test Webstorage by running ```yarn start``` from the ```examples/web``` folder. To use sessionStorage instead of localStorage, pass in ```true``` as the first parameter to ```WebStorage()``` in ```storage.js``` in the ```web``` folder. 
```const web = new WebStorage(true);```

There are still some things that should be done:
- Write tests
- Write documentation

Also, I am not very familiar with Typescript, so I expect that there may be some issues with how I implemented it.

### CODE REVIEW

**Comment by Krizzu** [comment]
@JesseRWeigel On the second thought, it's a lot to tackle at once: Storage, Example and Docs. Let's keep this PR with only Storage implementation and tests for it - let's leave other two for separate PRs.

**Comment by JesseRWeigel** [comment]
@Krizzu Good idea! I will keep everything in separate PRs.

**Comment by JesseRWeigel** [comment]
@Krizzu I removed the example app and finished the unit tests for local and session storage. Everything is working, but I have a few type errors that I haven't been able to resolve. Any assistance with the type errors that you could provide would be greatly appreciated.

```
packages/storage-web/__tests__/index.test.ts:34:35 - error TS2741: Property 'key2' is missing in type '{ key1: string; }' but required in type '{ key1: any; key2: any; }'.

34         await webStorage.setMany([{key1: 'value1'}, {key2: 'value2'}]);
                                     ~~~~~~~~~~~~~~~~

packages/storage-web/__tests__/index.test.ts:34:53 - error TS2741: Property 'key1' is missing in type '{ key2: string; }' but required in type '{ key1: any; key2: any; }'.

34         await webStorage.setMany([{key1: 'value1'}, {key2: 'value2'}]);
                                                       ~~~~~~~~~~~~~~~~

packages/storage-web/src/index.ts:53:5 - error TS2322: Type 'any[]' is not assignable to type '{ [k in K]: T[k] | null; }'.

53     return Promise.all(keys.map(k => this.storage.getItem(k)));
       ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~

packages/storage-web/src/index.ts:68:33 - error TS7053: Element implicitly has an 'any' type because expression of type 'string' can't be used to index type '{ [k in K]: T[k]; }'.
  No index signature with a parameter of type 'string' was found on type '{ [k in K]: T[k]; }'.

68       this.storage.setItem(key, keyValue[key]);
```

It seems like the first two type errors are related to the third one, and would be resolved along with it.

File: packages/storage-web/src/index.ts [hunk 1]
```diff
@@ -5,3 +5,98 @@
...
+    opts?: StorageOptions,
+  ): Promise<{[k in K]: T[k] | null}> {
+    if (opts) {
+      // noop
+    }
+    return Promise.all(keys.map(k => this.storage.getItem(k)));
+  }
+
+  async setMany<K extends keyof T>(
+    values: Array<{[k in K]: T[k]}>,
```
Krizzu:
```suggestion
    values: Array<Partial<{[k in K]: T[k]}>>,,
```

Otherwise, you'll be expected to pass full model every time

packages/storage-web/src/index.ts
```diff
@@ -5,3 +5,98 @@
...
+  }
+
+  async getMany<K extends keyof T>(
+    keys: Array<K>,
+    opts?: StorageOptions,
+  ): Promise<{[k in K]: T[k] | null}> {
+    if (opts) {
+      // noop
+    }
+    return Promise.all(keys.map(k => this.storage.getItem(k)));
```
Krizzu: This one is wrong - you're meant to return `Array<{key in keysOfModel}: T[key]}>`, instead, you're returning Array of values.

packages/core/src/defaults.ts
```diff
@@ -9,8 +9,8 @@
 import {FactoryOptions, LoggerAction} from '../types';
 
 export const factoryOptions: FactoryOptions = {
-  logger: __DEV__,
```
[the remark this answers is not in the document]
JesseRWeigel: My linter was complaining about it so I changed it just to be able to move forward. I will change it back.

packages/storage-web/__tests__/index.test.ts
```diff
@@ -0,0 +1,96 @@
...
+    });
+
+    it('removes all keys from localStorage', async () => {
+      localStorage.setItem('key1', 'value1');
+      localStorage.setItem('key2', 'value2');
+      await webStorage.dropStorage();
+      expect(localStorage.__STORE__).toEqual({});
+    });
+  });
+  // describe('utils', () => {
```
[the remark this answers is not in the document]
JesseRWeigel: yes

packages/storage-web/src/index.ts
JesseRWeigel: I am not sure exactly how to do this. I tried several things, but none of them worked. Maybe I am not exactly sure what is needed here.

Krizzu:
@JesseRWeigel Sorry for late response.

I've pushed a commit, that includes few changes to packages - this makes each of them an independent projects.

What's left to do here are typing of Storage.

packages/storage-web/src/index.ts
```diff
@@ -5,3 +5,94 @@
...
```
Krizzu: `getMany` is expected to return an object, where each key is one of provided `keys`, with value corresponding to its value in storage.

So you need to iterate over all keys, read them from storage and save value.

packages/storage-web/src/index.ts
```diff
@@ -5,3 +5,94 @@
...
+
+  async setMany<K extends keyof T>(
+    values: Array<Partial<{[k in K]: T[k]}>>,
+    opts?: StorageOptions,
+  ): Promise<void> {
+    if (opts) {
+      // noop
+    }
+    for (let keyValue of values) {
+      const key = Object.getOwnPropertyNames(keyValue)[0];
```
Krizzu:
```suggestion
      const key = Object.getOwnPropertyNames(keyValue)[0] as K;
```
You can help `typescript`, saying that `key` is indeed of type `K` (one of the keys in data object).

packages/storage-web/src/index.ts
```diff
@@ -5,3 +5,96 @@
...
+  }
+
+  async getMany<K extends keyof T>(
+    keys: Array<K>,
+    opts?: StorageOptions,
+  ): Promise<{[k in K]: T[k] | null}> {
+    if (opts) {
+      // noop
+    }
+    return Promise.all(keys.map(k => ({[k]: this.storage.getItem(k)}))).then(
```
JesseRWeigel: @Krizzu I got this to return an object containing the key value pair, but I can't figure out how to get the types working correctly. Please let me know what you think. Thanks for all your help and patience so far!

Krizzu: Sure thing.

Since this method (`getMany`) already returns a promise (because its marked as `async`), there's no need to return `Promise.all`. You can return the result of `reduce` directly.

So you'll eventually end up with something like this:

```typescript
   return keys.reduce(
      (storageValues, key) => {
        return {
          ...storageValues,
          [key]: this.storage.getItem(key),
        };
      },
      {} as {[k in K]: T[k] | null}, // here cast the initial value to the type we're going to return
    );
```

JesseRWeigel: Ah, this is much better! Thanks!

JesseRWeigel: Ok, I updated with your suggestion.

Krizzu:
@JesseRWeigel Great work, thanks a lot!

Issue #230, opened 2019-10-23
Add warning to iOS `setItem` if value is not stringified


Krizzu:
## Motivation

Values saved on iOS have to be stringified. We need a warning that will inform developer about it.

## Description

Add a check for platform and type of `value` before saving - if it's not a string, displaying warning.

jkossek:
@Krizzu  I'll take this

Krizzu:
Thanks, done!

PR #231, opened 2019-10-23
feat: add warning to set functions when value is not a string (#230)

jkossek:
Summary:
---------
Added warning on iOS when the value in the second argument of the `setItem` method is not a string.

Test Plan:
----------
Change the second argument of the `setItem` method in `example/src/examples/GetSetClear.js`, line 41 to non-string type and then try to increase the stored value. The warning will appear.

Example:
```
await AsyncStorage.setItem(STORAGE_KEY, newNumber);
await AsyncStorage.setItem(STORAGE_KEY, {newNumber});
await AsyncStorage.setItem(STORAGE_KEY, [newNumber]);
await AsyncStorage.setItem(STORAGE_KEY, null);
```

### DIFF
--- a/lib/AsyncStorage.js
+++ b/lib/AsyncStorage.js
@@ -11,7 +11,7 @@
 
 'use strict';
 
-const {NativeModules, Platform} = require('react-native');
+const {NativeModules} = require('react-native');
 
 const RCTAsyncStorage =
   NativeModules.RNC_AsyncSQLiteDBStorage ||
@@ -51,6 +51,14 @@ type MultiRequest = {|
   resolve: ?(result?: Promise<?$ReadOnlyArray<ReadOnlyArrayString>>) => void,
   reject: ?(error?: any) => void,
 |};
+
+function checkValueTypeNotString(value: any, usedKey: string) {
+  if (typeof value !== 'string') {
+    console.warn(
+      `[AsyncStorage] The value for key "${usedKey}" is not a string. This can lead to unexpected behavior/errors. Consider stringifying it.`,
+    );
+  }
+}
 
 /**
  * `AsyncStorage` is a simple, unencrypted, asynchronous, persistent, key-value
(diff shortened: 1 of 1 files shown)


lib/AsyncStorage.js
```diff
@@ -103,7 +103,19 @@ const AsyncStorage = {
         const errs = convertErrors(errors);
         callback && callback(errs && errs[0]);
         if (errs) {
-          reject(errs[0]);
+          if (Platform.OS !== 'ios') {
+            reject(errs[0]);
+          } else if (
+            errs[0] &&
+            String(errs[0]).match('Entries must be arrays of the form') &&
```
Krizzu: Turns out, the error `Entries must be arrays of the form` from native was shown when somehow wrongly parsed data was given. Since I upgraded the RN, it's no longer showing up.

We can revert those changes back to checking if value is not type of string, then displaying our warning.

jkossek: @Krizzu I've reverted it to the previous version.

lib/AsyncStorage.js
```diff
@@ -100,6 +100,11 @@ const AsyncStorage = {
   ): Promise<null> {
     return new Promise((resolve, reject) => {
       RCTAsyncStorage.multiSet([[key, value]], function(errors) {
+        if (Platform.OS === 'ios' && typeof value !== 'string') {
```
tido64: Any reason why we only do this on iOS?

lib/AsyncStorage.js
```diff
@@ -100,6 +100,11 @@ const AsyncStorage = {
...
+          console.warn(
```
tido64: There are a lot of completely unrelated changes in this PR. Can we split them out to separate PRs? The scope of this one should only contain changes to this file.

Krizzu: I had an issue to run the example app with Xcode 11, so I upgraded it. But good point, I'll create a different PR for that

lib/AsyncStorage.js
Krizzu: Just checked against android - this this warning should also be shown.

Krizzu: Also: on `multiSet` should also display warning

lib/AsyncStorage.js
```diff
@@ -301,6 +310,12 @@ const AsyncStorage = {
     callback?: ?(errors: ?$ReadOnlyArray<?Error>) => void,
   ): Promise<null> {
     return new Promise((resolve, reject) => {
+      keyValuePairs.forEach(([key, value]) => {
+        if (typeof value !== 'string') {
```
tido64: Should we move the check inside the new function as well?

Krizzu:
thanks @jkossek @tido64 !

PR #233, opened 2019-11-02
fix(example): make the file structure compatible with the react-nativ…

bartolkaruza:
This change makes the example compatible with the community [tester](https://github.com/react-native-community/tester).

Companion PR on the tester repo:
https://github.com/react-native-community/tester/pull/2


Krizzu:
@bartolkaruza I've applied those changes in #237, where I upgrade example RN version

PR #255, opened 2019-11-22
Prefer PlatformLocalStorage Over Other Native Modules

NickGerleman:
We currently try to load builtin native modules before trying to
fallback to platform provided native modules. Reversing this order
allows the platform to provide the AsyncStorage implementtion even if
on a platform that async-storage already supports. This ability can
be useful in hosted scenarios where we may want more control over how
data for an individual React Native instance is stored.

This should be entirely unneeded for AsyncStorageV2 since we have IStorageBackend.

### DIFF
--- a/lib/AsyncStorage.js
+++ b/lib/AsyncStorage.js
@@ -14,9 +14,9 @@
 const {NativeModules} = require('react-native');
 
 const RCTAsyncStorage =
+  NativeModules.PlatformLocalStorage || // Support for external modules, like react-native-windows
   NativeModules.RNC_AsyncSQLiteDBStorage ||
-  NativeModules.RNCAsyncStorage ||
-  NativeModules.PlatformLocalStorage; // Support for external modules, like react-native-windows
+  NativeModules.RNCAsyncStorage;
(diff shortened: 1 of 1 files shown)


Krizzu:
Hey @NickGerleman 

> This ability can
be useful in hosted scenarios where we may want more control over how
data for an individual React Native instance is stored.

Our initial thoughts were that `PlatformLocalStorage` module would be used by platforms that do not have the `RNCAsyncStorage` implementation. In case where both were present, `RNCAsyncStorage` is used anyway, leaving you no control over it. So I agree - this change looks good.

cc @tido64

NickGerleman:
Oh wow, I didn't actually realize that draft PRs were public. I originally made this just to run CI for a change we were playing with. I can still publish this if we like the idea of the change though.

NickGerleman:
@Krizzu @tido64 the review should be published now if you want to merge. It seems like the change could still be useful even if we ended up not needing it personally.

Krizzu:
@NickGerleman Thanks, going to release new version soon.

PR #259, opened 2019-11-25
fix(android): throw error when value is `null`

### LINKED ISSUE #253: Passing null or undefined as a value to setItem or multiSet doesn't result in an error.
## Current behavior

If `null` or `undefined` is passed as a value to `setItem` or `multiSet` then no error is generated and therefore any listening `.catch()` methods/`catch` blocks/error callbacks don't fire to let you know something went wrong.

## Expected behavior

I would expect an error to be generated in the same way as if any other invalid value was passed e.g. number, bool, object.

## Repro steps

```
            try {
              await AsyncStorage.setItem('aTestKey', null, e => {
                console.error(`Error in callback`, { e });
              });
            } catch (e) {
              console.error(`Error in catch`, { e });
            }
```

## Environment

- **Async Storage version**: 1.6.2
- **React-Native version**: 0.61.4
- **Platform tested**: Android
- **Logs/Error that are relevant**: N/A

Krizzu:
Summary:
---------
 
Android did not show errors when `null`/`undefined` values were passed. This PR fixes that.

Closes #253 

Test Plan:
----------

1. Try setting a  `null`/`undefined` value to AsyncStorage.
2. Note that errors where thrown.

### DIFF
--- a/lib/AsyncStorage.js
+++ b/lib/AsyncStorage.js
@@ -52,11 +52,25 @@ type MultiRequest = {|
   reject: ?(error?: any) => void,
 |};
 
-function checkValueTypeNotString(value: any, usedKey: string) {
-  if (typeof value !== 'string') {
+function checkValidInput(usedKey: string, value: any) {
+  const isValuePassed = arguments.length > 1;
+
+  if (typeof usedKey !== 'string') {
     console.warn(
-      `[AsyncStorage] The value for key "${usedKey}" is not a string. This can lead to unexpected behavior/errors. Consider stringifying it.`,
+      `[AsyncStorage] Using ${typeof usedKey} type for key is not supported. This can lead to unexpected behavior/errors. Use string instead.\nKey passed: ${usedKey}\n`,
     );
+  }
+
+  if (isValuePassed && typeof value !== 'string') {
+    if (value == null) {
+      throw new Error(
+        `[AsyncStorage] Passing null/undefined as value is not supported. If you want to remove value, Use .remove method instead.\nPassed value: ${value}\nPassed key: ${usedKey}\n`,
+      );
+    } else {
+      console.warn(
+        `[AsyncStorage] The value for key "${usedKey}" is not a string. This can lead to unexpected behavior/errors. Consider stringifying it.\nPassed value: ${value}\nPassed key: ${usedKey}\n`,
+      );
+    }
   }
 }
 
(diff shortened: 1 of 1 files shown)


android/src/main/java/com/reactnativecommunity/asyncstorage/AsyncStorageModule.java
```diff
@@ -218,23 +218,21 @@ protected void doInBackgroundGuarded(Void... params) {
           for (int idx=0; idx < keyValueArray.size(); idx++) {
             if (keyValueArray.getArray(idx).size() != 2) {
               error = AsyncStorageErrorUtil.getInvalidValueError(null);
-              return;
```
tido64: Won't we risk overwriting errors now? Should they be concatenated? Should we throw instead?

android/src/main/java/com/reactnativecommunity/asyncstorage/AsyncStorageModule.java
```diff
@@ -218,23 +218,21 @@ protected void doInBackgroundGuarded(Void... params) {
...
             if (keyValueArray.getArray(idx).getString(1) == null) {
               error = AsyncStorageErrorUtil.getInvalidValueError(null);
-              return;
             }
-
-            statement.clearBindings();
-            statement.bindString(1, keyValueArray.getArray(idx).getString(0));
-            statement.bindString(2, keyValueArray.getArray(idx).getString(1));
-            statement.execute();
+            if(error == null) {
```
tido64: Nit: Space after `if`

Krizzu:
@tido64 Checks are not done in get/set functions, where passing `null` or `undefined` as value throws an error from JS (it would be thrown from Native either way).

lib/AsyncStorage.js
```diff
@@ -52,10 +52,24 @@ type MultiRequest = {|
...
+function checkValidInput(usedKey: string, value: any) {
+  const isValuePassed = arguments.length > 1;
+
+  if (typeof usedKey !== 'string') {
+    console.warn(
+      `[AsyncStorage] Using ${typeof usedKey} type for key is not supported. This can lead to unexpected behavior/errors. Use string instead.\nKey passed: ${usedKey}\n`,
+    );
+  }
+
+  if (isValuePassed && value == null) {
```
tido64: Should we nest this check inside the if-statement below?

lib/AsyncStorage.js
```diff
@@ -336,6 +353,10 @@ const AsyncStorage = {
     callback?: ?(errors: ?$ReadOnlyArray<?Error>) => void,
   ): Promise<null> {
     return new Promise((resolve, reject) => {
+      keys.forEach(key => {
```
tido64: This can be simplified:
```js
keys.forEach(checkValidInput);
```

lib/AsyncStorage.js
Krizzu: Yea, good catch

tido64:
Thanks, @Krizzu ❤️

PR #261, opened 2019-11-30
fix(iOS): Remove RNC -> RCT storage migration

tido64:
Summary:
---------

Removed migration of `RNCAsyncLocalStorage_V1` to `RCTAsyncLocalStorage_V1` in preparation for AsyncStorage v2.

Test Plan:
----------

All current tests should pass.


Krizzu:
Thanks!

PR #268, opened 2019-12-13
Update AsyncStorage.js

retyui:
Summary:
---------
```js

AsyncStorage.multiRemove(['a']);

/*
[AsyncStorage] The value for key "a" is not a string. This can lead to unexpected behavior/errors. Consider stringifying it.
Passed value: 0
Passed key: a
*/
```

Fixed https://github.com/react-native-community/async-storage/issues/267

Test Plan:
----------
```js
AsyncStorage.multiRemove(['a']);
```

### DIFF
--- a/lib/AsyncStorage.js
+++ b/lib/AsyncStorage.js
@@ -353,7 +353,7 @@ const AsyncStorage = {
     callback?: ?(errors: ?$ReadOnlyArray<?Error>) => void,
   ): Promise<null> {
     return new Promise((resolve, reject) => {
-      keys.forEach(key => checkValidInput(key, 'value'));
+      keys.forEach(key => checkValidInput(key));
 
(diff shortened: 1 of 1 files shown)


lib/AsyncStorage.js
```diff
@@ -353,7 +353,7 @@ const AsyncStorage = {
     callback?: ?(errors: ?$ReadOnlyArray<?Error>) => void,
   ): Promise<null> {
     return new Promise((resolve, reject) => {
-      keys.forEach(checkValidInput);
+      keys.forEach(key => checkValidInput(key, 'value'));
```
Krizzu:
```suggestion
      keys.forEach(key => checkValidInput(key));
```

You don't have to pass a second argument here

Krizzu:
Thanks!

PR #270, opened 2019-12-27
Don't reset the storage when the manifest file cannot be read because of iOS data protection

aleh:
Summary:
---------

This is to distinguish "manifest file does not exist or is corrupted" and "have no access to the manifest file yet" cases.

Starting with an empty dictionary is a reasonable fallback in case the manifest does not exist or cannot be parsed because of being corrupted or incompatible. However doing so might lead to unnecessary data loss when the app simply has no access to its documents yet due to iOS data protection (if enabled and when the app launches while the device is locked due to e.g. geofencing, silent notification, external accessory, etc).

Test Plan:
----------

Enable data protection in your RN iOS app, then trigger a geofencing event or a silent push notification to launch the app while the phone is locked and files are not readable. If the app tries to access async storage in this case, then it'll end up erasing it without this fix.

### DIFF
--- a/ios/RNCAsyncStorage.m
+++ b/ios/RNCAsyncStorage.m
@@ -348,20 +348,20 @@ RCT_EXPORT_MODULE()
     NSDictionary *errorOut = nil;
     NSString *serialized = RCTReadFile(RCTGetManifestFilePath(), RCTManifestFileName, &errorOut);
     if (!serialized) {
-    	if (errorOut) {
-     		// We cannot simply create a new manifest in case the file does exist but we have no access to it.
-				// This can happen when data protection is enabled for the app and we are trying to read the manifect
-				// while the device is locked. (The app can be started by the system even if the device is locked due to
-				// e.g. a geofence event.)
-      	RCTLogError(@"Could not open the existing manifest, perhaps data protection is enabled?\n\n%@", errorOut);
-      	return errorOut;
+      if (errorOut) {
+        // We cannot simply create a new manifest in case the file does exist but we have no access to it.
+        // This can happen when data protection is enabled for the app and we are trying to read the manifect
+        // while the device is locked. (The app can be started by the system even if the device is locked due to
+        // e.g. a geofence event.)
+        RCTLogError(@"Could not open the existing manifest, perhaps data protection is enabled?\n\n%@", errorOut);
+        return errorOut;
       } else {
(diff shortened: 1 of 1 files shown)


Krizzu:
@aleh  Thanks for working on that. I'll have a look as soon as I can.
cc @tido64

ios/RNCAsyncStorage.m
```diff
@@ -343,16 +343,33 @@ - (NSDictionary *)_ensureSetup
...
-    NSDictionary *errorOut;
+    NSDictionary *errorOut = nil;
     NSString *serialized = RCTReadFile(RCTGetManifestFilePath(), RCTManifestFileName, &errorOut);
-    _manifest = serialized ? RCTJSONParseMutable(serialized, &error) : [NSMutableDictionary new];
-    if (error) {
-      RCTLogWarn(@"Failed to parse manifest - creating new one.\n\n%@", error);
-      _manifest = [NSMutableDictionary new];
+    if (!serialized) {
+      if (errorOut) {
+        // We cannot simply create a new manifest in case the file does exist but we have no access to it.
```
tido64: Can we be confident that this is always the case if we get an error? Can we check whether the file exists first?

aleh: This is what RCTReadFile() is doing — errorOut is nil in case the file does not exist.

tido64: IMO, it is safer to do an existence check before we read the file rather than relying on the behavior of a private API to stay consistent.

WDYT, @Krizzu?

aleh: `RCTReadFile()` is our method defined on line 59 above, we can rely on its behavior.

tido64: Ah, you're absolutely right. I'm blind!

ios/RNCAsyncStorage.m
```diff
@@ -343,16 +343,33 @@ - (NSDictionary *)_ensureSetup
...
+        // This can happen when data protection is enabled for the app and we are trying to read the manifect
```
Krizzu:
```suggestion
        // This can happen when data protection is enabled for the app and we are trying to read the manifest
```

Small typo

aleh: Thanks!

PR #303, opened 2020-02-19
Create [bundle ID] folder during migration if none exists

frankenthumbs:
Summary:
---------

fixes issue described in https://github.com/react-native-community/async-storage/issues/301

Test Plan:
----------

you can use my sample repo here: https://github.com/frankenthumbs/AsyncStorageUpgradePOC and use the repro steps but the gist is, on iOS:
1. save something in async storage on version 1.2.1
2. upgrade to latest
3. noticed data doesn't get wiped

### DIFF
--- a/ios/RNCAsyncStorage.m
+++ b/ios/RNCAsyncStorage.m
@@ -213,11 +213,29 @@ static void RCTStorageDirectoryCleanupOld(NSString *oldDirectoryPath)
   }
 }
 
+static void _createStorageDirectory(NSString *storageDirectory, NSError **error)
+{
+  [[NSFileManager defaultManager] createDirectoryAtPath:storageDirectory
+                            withIntermediateDirectories:YES
+                                             attributes:nil
+                                                  error:error];
+}
+
 static void RCTStorageDirectoryMigrate(NSString *oldDirectoryPath, NSString *newDirectoryPath, BOOL shouldCleanupOldDirectory)
 {
   NSError *error;
   // Migrate data by copying old storage directory to new storage directory location
   if (![[NSFileManager defaultManager] copyItemAtPath:oldDirectoryPath toPath:newDirectoryPath error:&error]) {
+    // the new storage directory "Application Support/[bundleID]/RCTAsyncLocalStorage_V1" seems unable to migrate
+    // because folder "Application Support/[bundleID]" doesn't exist.. create this folder so and attempt to migrate again
+    if (error != nil && error.code == 4 && [newDirectoryPath hasSuffix:RCTStorageDirectory]) {
+        NSError *error = nil;
+        _createStorageDirectory(RCTCreateStorageDirectoryPath(@""), &error);
+        if (error == nil) {
+          RCTStorageDirectoryMigrate(oldDirectoryPath, newDirectoryPath, shouldCleanupOldDirectory);
+          return;
+        }
+      }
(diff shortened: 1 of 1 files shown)


grit96:
Please can we get this merged? Currently we can't update to 1.8.0 due to the failed migration.

ios/RNCAsyncStorage.m
```diff
@@ -213,11 +213,29 @@ static void RCTStorageDirectoryCleanupOld(NSString *oldDirectoryPath)
...
   if (![[NSFileManager defaultManager] copyItemAtPath:oldDirectoryPath toPath:newDirectoryPath error:&error]) {
+    // the new storage directory "Application Support/[bundleID]/RCTAsyncLocalStorage_V1" seems unable to migrate
+    // because folder "Application Support/[bundleID]" doesn't exist.. create this folder and attempt folder copying again
+    if (error != nil && error.code == 4 && [newDirectoryPath hasSuffix:RCTStorageDirectory]) {
+        NSError *error = nil;
+        _createStorageDirectory(RCTCreateStorageDirectoryPath(@""), &error);
+        if (error == nil) {
+          RCTStorageDirectoryMigrate(oldDirectoryPath, newDirectoryPath, shouldCleanupOldDirectory);
+          return;
+        }
```
Krizzu:
```suggestion
        } else {
          RCTStorageDirectoryMigrationLogError(@"Failed to create storage directory during the migration.", error);
          return;
        }
```

Let's warn about directory creation fail

JerakRus:
Очень ждем этот фикс, спасибо!

Krizzu:
@HeyImChris Hey, would you mind having a look here?

ios/RNCAsyncStorage.m
```diff
@@ -213,12 +213,32 @@ static void RCTStorageDirectoryCleanupOld(NSString *oldDirectoryPath)
...
+                            withIntermediateDirectories:YES
+                                             attributes:nil
+                                                  error:error];
+}
+
 static void RCTStorageDirectoryMigrate(NSString *oldDirectoryPath, NSString *newDirectoryPath, BOOL shouldCleanupOldDirectory)
 {
   NSError *error;
   // Migrate data by copying old storage directory to new storage directory location
   if (![[NSFileManager defaultManager] copyItemAtPath:oldDirectoryPath toPath:newDirectoryPath error:&error]) {
```
HeyImChris: What if we put a fileExistsAtPath: check before this and if false, call _createStorageDirectory. Then add an "error != nil" to the copyItemAtPath: check to make sure the fileExistsAtPath succeeded and leave the rest of the method unchanged?

Seems a little messy to _try_ to copy and recursively respond to that failure instead of checking if we should create first, putting ourselves in a consistent state in all cases and going from there.

frankenthumbs: hey chris.. i'm down to make this change but i was reading on the apple docs that its better to attempt an operation and respond than to try to predict on how an operation will go based on the file system's current state. https://developer.apple.com/documentation/foundation/nsfilemanager/1415645-fileexistsatpath

since this is a race condition issue and we are the only thing writing to that folder, it'll be okay.
i'll test and try to get something up within today

HeyImChris: Didn't know about that doc. I think what you're doing is good then!

Krizzu:
Thanks a lot everyone!

PR #306, opened 2020-02-25
update react native doc links with new docs url

bell-steven:
Summary:
---------

Facebook changed the URL of the React Native docs from `facebook.github.io/react-native` to `reactnative.dev`.  This PR updated all the links pointing to the old docs site with the updated link. 

Test Plan:
----------

No code changed. Only comments were updated.


Krizzu:
Thanks @bell-steven

PR #309, opened 2020-03-03
update Error message

ahoseinian:
Just a one-line error message change because there is no `--clearCache` flag on packager.


Krizzu:
Thanks

PR #313, opened 2020-03-15
Removed conflict in type for create static method and made opts param…

ocarlott:
Summary:
---------

I removed conflict in type for create static method and made opts parameter optional

Test Plan:
----------

No new code introduced other than fixing typescript typings. Passed all existing tests.


Krizzu:
Thanks!

PR #327, opened 2020-04-06
Add Windows AsyncStorage

kaiguo:
Summary:
---------

Add Windows AsyncStorage.
Closes https://github.com/microsoft/react-native-windows/issues/2271.

Test Plan:
----------

0. [System requirments](https://microsoft.github.io/react-native-windows/docs/rnw-dependencies).
1. Run `yarn start:windows`.
2. Open `example/windows/AsyncStorageExample.sln` with VisualStudio and run.

Screenshot_20200406162637


kaiguo:
@Krizzu there are bunch of Flow check errors inside `react-native/Libraries`, any idea how to address them?

https://circleci.com/gh/react-native-community/async-storage/2125?utm_campaign=vcs-integration-link&utm_medium=referral&utm_source=github-build-link

windows/ReactNativeAsyncStorage/RNCAsyncStorage.h
```diff
@@ -0,0 +1,152 @@
...
+                            callback(errors, {});
+                        }
+                    }
+                });
+        }
+
+        REACT_METHOD(multiSet);
+        void multiSet(std::vector<JSValue>& pairs, std::function<void(JSValueArray const&)>&& callback) noexcept {
+            dbStorage.AddTask(DBStorage::DBTask::Type::multiSet, std::move(pairs),
+                [callback](std::vector<JSValue> const& callbackParams) {
```
htpiv, line 36: Move `callback` into the closure?

windows/ReactNativeAsyncStorage/RNCAsyncStorage.h
```diff
@@ -0,0 +1,152 @@
...
+{
+    REACT_MODULE(RNCAsyncStorage);
+    struct RNCAsyncStorage
+    {
+        DBStorage dbStorage;
+
+        REACT_METHOD(multiGet);
+        void multiGet(std::vector<JSValue>& keys, std::function<void(JSValueArray const& errors, JSValueArray const& results)>&& callback) noexcept {
+            dbStorage.AddTask(DBStorage::DBTask::Type::multiGet, std::move(keys),
+                [callback](std::vector<JSValue> const& callbackParams) {
```
htpiv, line 20: Move `callback` into the closure?

windows/ReactNativeAsyncStorage/DBStorage.cpp
```diff
@@ -0,0 +1,451 @@
...
+    {
+        // If there is an in-progress async task, cancel it and wait on the
+        // condition_variable for the async task to acknowledge cancellation by
+        // nulling out m_action. Once m_action is null, it is safe to proceed
+        // wth closing the DB connection
+        winrt::slim_lock_guard guard{ m_lock };
+        m_tasks.clear();
+        if (m_action) {
+            m_action.Cancel();
+            m_cv.wait(m_lock, [this]() { return m_action == nullptr; });
```
htpiv, line 243: Do we know what kind of apartment that `DBStorage`'s destructor will run in? We're blocking here without pumping messages, which could lead to deadlocks in general for a STA or ASTA thread (for a MTA thread, it's just impolite. 🙂)

Krizzu:
@kaiguo It's caused by bumping RN version, while Flow stays on the old one.
Is there a reason for changing it to 0.61? I haven't tested `AsyncStorage` with that version yet.

windows/ReactNativeAsyncStorage/RNCAsyncStorage.h
kaiguo: You mean make callback a parameter of the lambda?

I wasn't able to since callback types are different. `CxxModule::Callback` is not available here. If you look closely, some callbacks are `std::function<void(JSValueArray const&)>`, others are `std::function<void(JSValueArray const& errors, JSValueArray const& results)>`. We can make `AddTask` take different parameters but then we moved the burden to `DBTask`, it has to take account all the different callback types, which feels more complicated than just capturing the callback here.

windows/ReactNativeAsyncStorage/DBStorage.cpp
```diff
@@ -0,0 +1,451 @@
...
+        const Statement& stmt,
+        int index,
+        const std::string& str) {
+        return CheckSQLiteResult(db, callback, sqlite3_bind_text(stmt.get(), index, str.c_str(), -1, SQLITE_TRANSIENT));
+    }
+
+} // namespace
+
+DBStorage::DBStorage() {
+    auto const localAppDataParh = winrt::ApplicationData::Current().LocalFolder().Path();
```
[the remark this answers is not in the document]
kaiguo: I'll check for package identity before using it. Where should we store the file if this path is not available? I don't think UWP app has the permission to write to arbitrary locations?

windows/ReactNativeAsyncStorage/DBStorage.cpp
kaiguo: I'm a little bit fuzzy here so correct me if I'm wrong...

We used it as a stack variable in RNCAsyncStorage, so the destructor should be called when the AsyncStorage module gets destructed. DBStorage is not a public type so outside users won't be able to touch it. So I *think* we are good?

kaiguo:
> Is there a reason for changing it to 0.61? I haven't tested `AsyncStorage` with that version yet.

@Krizzu `react-native-windows` has a peer dependency on RN 0.61.5. I guess I can revert it to 0.58 for now, the downside is windows sample won't compile/run😢.

windows/ReactNativeAsyncStorage/RNCAsyncStorage.h
htpiv: No, I mean something like, say:
```c++
[cb{std::move(callback)}](std::vector<JSValue> const& callbackParams) {
  ...
  cb(...);
  ...
}
```

kaiguo: Oh I see, will update. Thanks for the example!

Krizzu:
> @Krizzu `react-native-windows` has a peer dependency on RN 0.61.5. I guess I can revert it to 0.58 for now, the downside is windows sample won't compile/run😢.

I see. Well, we just need to make sure that AS works with 0.61.5 now. 
For RN 0.61.5, try upgrading to Flow version 0.105.0. Here's a [config used in RN 0.61.5](https://raw.githubusercontent.com/react-native-community/rn-diff-purge/release/0.61.5/RnDiffApp/.flowconfig) - I guess we just need to do some adjustments.

windows/ReactNativeAsyncStorage/DBStorage.cpp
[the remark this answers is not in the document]
htpiv: I think the DB name should be configurable at either build-time or run-time. If you want to have it be configured at runtime, you could look for a string in `Windows.ApplicationMode.Core.CoreApplication.Properties`... say, `React-Native-Community-Async-Storage-Database-Path`? Verbose enough? :) On UWP, if there's no such value, you'd fallback to `ApplicationData`. For Win32, you'd want to error out.

Build-time configuration would be simpler, but it would preclude distributing asyncStorage as a DLL. I forget if that's a goal or not...

windows/ReactNativeAsyncStorage/DBStorage.cpp
htpiv: That sounds right, but it just shifts the questions up a level... what kind of thread is the `AsyncStorage` module being destroyed on?

kaiguo: I changed it to shared lock like you did in the win32 module, which I think should avoid the deadlock.

kaiguo:
@Krizzu flowconfig changes did fix the flow check errors, but looks like both iOS and Android builds are broken after upgrading to 61. I created a separated branch [here](https://github.com/kaiguo/async-storage/tree/upgrade-sample-app-to-61) which fixed Android and Flow issues. For iOS, I think we need to switch to pod linking, looks like that's the only supported way on 61+. I'm not very familiar with iOS so I created https://github.com/react-native-community/async-storage/issues/328 to track the rest of the work.

I'll keep react-native on 59 in this PR for now.

kaiguo:
PR that upgrades iOS and Android to RN 61: https://github.com/react-native-community/async-storage/pull/329.

Krizzu:
@kaiguo Great, let's fix merge #328 first, so you can keep this PR with RN 61.

windows/ReactNativeAsyncStorage/DBStorage.cpp
```diff
@@ -194,13 +195,45 @@ namespace {
...
+            m_mutex.unlock_shared();
+        }
+
+    private:
+        winrt::slim_mutex& m_mutex;
+    };
+
 } // namespace
 
+const winrt::hstring DBStorage::s_dbPathProperty = L"React-Native-Community-Async-Storage-Database-Path";
```
htpiv, line 217: This does a spurious dynalloc at boot, and destroys the dynamically allocated memory in dllmain, which is not good. I would recommend instead saying something like:
```c++
constexpr auto dbPathProperty = L"React-Native-Community-Async-Storage-Database-Path";
if (auto pathInspectable = winrt::CoreApplication::Properties().TryLookup(dbPathProperty)) {
...
```
Which will use a fast-pass `HSTRING` and avoids any dynamic allocations. The microoptimization of avoid 1 more string alloc is not a big deal, but avoiding the calls to `WindowsCreateString` and `WindowsDeleteString` from inside of `DllMain` is good.

windows/ReactNativeAsyncStorage/DBStorage.cpp
htpiv: Hmm, no, that's not what I did in the win32 module - ping me on Teams tomorrow during working hours and I'll explain my question here a little more fully.

Krizzu:
@kaiguo Sorry for keeping you waiting, but I'm waiting for someone else to have a look here (with Windows experience, asked at RN channel).

kaiguo:
> @kaiguo Kai Guo FTE Sorry for keeping you waiting, but I'm waiting for someone else to have a look here (with Windows experience, asked at RN channel).

Sure, good to have some extra eyes on this : )

package.json
```diff
@@ -48,9 +49,9 @@
     "@babel/core": "^7.6.2",
     "@babel/runtime": "^7.6.2",
     "@react-native-community/eslint-config": "^0.0.5",
-    "@react-native-community/cli": "4.6.3",
-    "@react-native-community/cli-platform-ios": "4.6.3",
-    "@react-native-community/cli-platform-android": "4.6.3",
+    "@react-native-community/cli": "^3.1.0",
```
empyrical, line 52: Are there issues with the the newer version of RNCLI?

kaiguo: 4.x has a peer dependency on RN 62, the test apps are on 61 so it generated bunch of warnings in `yarn install`. 3.x is probably a better match for RN 61.

Krizzu:
I think there's no need to prolong this PR. Thanks for all the work @kaiguo 💪 🎉

kaiguo:
@Krizzu thanks for merging this in.

I'm trying to use the module in another project and just noticed the changes haven't been published yet, do we know when the next release will be?

Krizzu:
@kaiguo I'm going to issue new release later today

Issue #349, opened 2020-05-09
V2: Outdated typings


maximzhemerenko:
Hi Guys, thanks a lot, I found v2 version of this library very useful for me, but noticed one problem.

## Current behavior

Typescript typings are outdated with the current implementation.
For example, WebStorage class should have constructor, but typings not...

## Expected behavior

Typings should be up to date.

## Repro steps

Write a basic sample from the Readme file using typescript. Compilator says `TS2554: Expected 0 arguments, but got 1.`.

## Environment

"@react-native-community/async-storage-backend-web": "^2.0.0-rc.3"

- **React-Native version**: - Not react native, "react": "^16.13.1",
- **Platform tested**: - Web platform
- **Logs/Error that are relevant**: TS2554: Expected 0 arguments, but got 1.

Who can help me, what can I do to find a workaround to this issue? I can help and make a merge request, but I don't know how to generate typing in a proper way? And how I can test is my fixes works fine?
I noticed that someone removed `yarn generate:types` in #216, but I do not completely understand how it does works)

tido64:
No longer applicable.

PR #350, opened 2020-05-09
Update index.d.ts files for storage-legacy and storage-web packages

maximzhemerenko:
Summary:
---------

This MR related to issue #349. I've updated index.d.ts files using tsc compiler manually. I did not understand how d.ts files should be generated now, so please suggest me to find the best way to automate generating d.ts files to avoid this problem in the future. Thanks.

Test Plan:
----------

After this change, the bug should diapers.
Please see issue #349, I described how to reproduce the bug.

### DIFF
--- a/packages/storage-legacy/types/index.d.ts
+++ b/packages/storage-legacy/types/index.d.ts
@@ -5,42 +5,20 @@
  * LICENSE file in the root directory of this source tree.
  *
  */
-
-import {
-  EmptyStorageModel,
-  IStorageBackend,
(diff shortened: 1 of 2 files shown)


Krizzu:
Hey, been a while.

It seems like there's small issue with formatting - would you have a look?

maximzhemerenko:
As I said I generated types file using TS-compiler, not manually.

Krizzu:
I get you, but still, this could use some formatting. Can you run `eslint` through that file?

PR #351, opened 2020-05-11
fix: Web compatibility

Krizzu:
Summary:
---------

Since we target web, Async Storage needs to be available to use by CRA too.
What has been done:

- moved source to `src/`
- used [bob](https://github.com/react-native-community/bob) to create `commonjs`/`es modules` compatible bundles
- Changed [mock](https://github.com/react-native-community/async-storage/blob/master/docs/Jest-integration.md) to be `commonjs` module and type-less (removed Flow)

Test Plan:
----------

Green CI.


package.json
```diff
@@ -92,5 +108,13 @@
         "name": "Emu_E2E"
       }
     }
+  },
+  "@react-native-community/bob": {
+    "source": "src",
+    "output": "lib",
+    "targets": [
+      "commonjs",
```
satya164, line 116: Since you're using Flow, I'd suggest to use the `copyFlow` option as well so that flow definitions are copied

```suggestion
      ["commonjs", {"copyFlow": true}],
```

package.json
```diff
@@ -92,5 +108,13 @@
...
+      "module"
```
satya164, line 117:
```suggestion
      ["module", {"copyFlow": true}],
```

package.json
Krizzu: Makes sense, thanks :pray:

PR #359, opened 2020-05-15
Fix Windows project imports

kaiguo:
Summary:
---------

Switched to `$(ReactNativeWindowsDir)` variable instead of searching for the hard coded path, this unblocks our uses cases inside react-native-windows repo.

Test Plan:
----------

Run the Windows example app as usual.


kaiguo:
Some minor fixes, @Krizzu could you help get this merged and released? Thanks!

I've also removed the .gitignore file in bundle folder, hopefully this will stop making the bundle folder being included in the release package.

PR #368, opened 2020-05-22
Swap the native module used when in Expo managed workflow

brentvatne:
Summary:
---------

Currently tools like AWS Amplify and redux-persist need separate instructions for using AsyncStorage whether you are in an Expo managed project or a bare React Native project, because Expo manage projects do not yet include @react-native-community/async-storage. We would like to include it, but it's a bit trickier than most packages due to complexity around scoping storage access when switching between different projects within the Expo client app.

In order to let library authors just tell people to use @react-native-community/async-storage, I think the easiest solution for now is to swap out the backing native module when in the Expo managed workflow to the solution provided by React Native, which is still included in core as recently as 0.62.2. I did this by moving out the logic for picking which native module to use into `RCTAsyncStorage.js` and `RCTAsyncStorage.expo.js`. The `.expo.js` extension is only used in the Expo managed workflow ([more info if you're curious](https://docs.expo.io/bare/using-expo-client/#use-the-expojsjsontstsx-extension-to-provide-expo)).

As far as I can tell, the API interface is the same and this should not cause any issues. When we update to 0.63 we will likely investigate what it will take to fully integrate this library.

**The changes here have no impact on any environment other than Expo managed projects.**

Test Plan:
----------

To make this very easy to test I published an alpha version of the package: `@react-native-community/async-storage@1.11.0-alpha.0`.

You can test that this works in Expo managed apps by running the following:
- `npx crna --template blank`
- Install `@react-native-community/async-storage@1.11.0-alpha.0`
- Copy in the ["useAsyncStorage Specific Example"](https://github.com/react-native-community/async-storage/blob/master/docs/API.md#useasyncstorage) into App.js
- `yarn ios` and/or `yarn android`
- Import it in the app and use it

And in a bare React Native app:
- `npx react-native init`
- Install `@react-native-community/async-storage@1.11.0-alpha.0`
- `npx pod-install`
- Copy in the ["useAsyncStorage Specific Example"](https://github.com/react-native-community/async-storage/blob/master/docs/API.md#useasyncstorage) into App.js
- `yarn ios` and/or `yarn android`
- Import it in the app and use it

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@react-native-community/async-storage",
-  "version": "1.11.0-alpha.0",
+  "version": "1.10.3",
   "description": "Asynchronous, persistent, key-value storage system for React Native.",
   "main": "lib/commonjs/index.js",
   "module": "lib/module/index.js",
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@react-native-community/async-storage",
-  "version": "1.10.3",
+  "version": "1.11.0-alpha.0",
   "description": "Asynchronous, persistent, key-value storage system for React Native.",
   "main": "lib/commonjs/index.js",
   "module": "lib/module/index.js",
--- a/.github/CODEOWNERS
+++ b/.github/CODEOWNERS
@@ -1,0 +1,2 @@
+# Expo managed workflow supported native backend
+/src/RCTAsyncStorage.expo.js @brentvatne
--- a/src/AsyncStorage.native.js
+++ b/src/AsyncStorage.native.js
@@ -11,12 +11,7 @@
 
 'use strict';
 
-const {NativeModules} = require('react-native');
-
-const RCTAsyncStorage =
-  NativeModules.PlatformLocalStorage || // Support for external modules, like react-native-windows
-  NativeModules.RNC_AsyncSQLiteDBStorage ||
-  NativeModules.RNCAsyncStorage;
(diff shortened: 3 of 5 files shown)


Ashish-Nanda:
This is a major issue for libraries like AWS Amplify as @brentvatne stated above, where we try to support Expo managed apps as well as bare apps created by the RNC CLI. Currently we have held off on migrating to the community package for this reason, but it is one of the most frequent requests from customers on our repo. You can see a recent comment where I've addresed it on a customer PR here: https://github.com/aws-amplify/amplify-js/pull/4027#issuecomment-628050861
 
It is also a major pain point when an app relies on libraries some of which use AsyncStorage from Core and others that rely on `@react-native-community/async-storage` because it leads to a race condition https://github.com/react-native-community/async-storage/issues/118#issue-452790369. This issue was opened, incidentally by an Amplify customer and responded to by @Krizzu who root caused it.

I think this PR is probably the fastest path forward in enabling Expo managed apps to support `@react-native-community/async-storage` and this would be a great way to unblock customers and library developers while not impacting the existing functionality in any other kind of React Native apps.

package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@react-native-community/async-storage",
-  "version": "1.10.3",
+  "version": "1.11.0-alpha.0",
```
Krizzu, line 3: Please reverse it

brentvatne: done!

PR #400, opened 2020-07-11
Fix API typos

TheDSCPL:
Summary:
---------

Fix API typos

Test Plan:
----------
No test required

### DIFF
--- a/website/docs/API.md
+++ b/website/docs/API.md
@@ -25,7 +25,7 @@ static getItem(key: string, [callback]: ?(error: ?Error, result: ?string) => voi
 
 `Promise` resolving with a string value, if entry exists for given `key`, or `null` otherwise.
 
-`Promise` rejects in case of underlying storage error.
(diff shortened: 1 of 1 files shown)


Krizzu:
Can you explain why you think this is a typo?

TheDSCPL:
"can be also reject**s**"? It looks like the author wanted to type "can be also reject**ed**". I didn't change to that because "can be also" is totally unnecessary due to "in case of". If you really want the "can also be" to stay, then it should at least be "can be also reject**ed**"

TheDSCPL:
On second thought it actually should be "rejecting in case(...)" to match the tense from the previous line which starts with "resolving"

TheDSCPL:
I updated the PR to "can also be rejected" for middle ground :)

Krizzu:
> I updated the PR to "can also be rejected" for middle ground :)

Appreciated 🙏

PR #418, opened 2020-08-12
Add support for enabled TurboModules in Expo

sjchmiela:
Summary:
---------

At Expo we're on the verge of enabling TurboModules in SDK 39 (https://github.com/expo/expo/pull/9687). While we haven't added the native code for this module to the SDK yet, we want to maintain the `.expo.js` fallback, so it resolves with the React Native's `AsyncStorage`.

Test Plan:
----------

I have verified manually that copying and pasting this code to an appropriate file in `node_modules` for a running Expo project fixes the `[@RNC/AsyncStorage]: NativeModule: AsyncStorage is null.` error.

The `if (TurboModuleRegistry)` helps to not break this code on older versions of RN which do not have TMR (I hope).

### DIFF
--- a/src/RCTAsyncStorage.expo.js
+++ b/src/RCTAsyncStorage.expo.js
@@ -7,7 +7,18 @@
  * will likely not be valid anymore, and the package will need to be included in the Expo SDK
  * to continue to work.
  */
-const {NativeModules} = require('react-native');
-const RCTAsyncStorage = NativeModules.AsyncSQLiteDBStorage || NativeModules.AsyncLocalStorage;
+const {NativeModules, TurboModuleRegistry} = require('react-native');
+
+let RCTAsyncStorage;
+
+// TurboModuleRegistry falls back to NativeModules
+// so we don't have to try go assign NativeModules'
+// counterparts if TurboModuleRegistry would resolve
+// with undefined.
+if (TurboModuleRegistry) {
+  RCTAsyncStorage = TurboModuleRegistry.get("AsyncSQLiteDBStorage") || TurboModuleRegistry.get("AsyncLocalStorage");
+} else {
+  RCTAsyncStorage = NativeModules.AsyncSQLiteDBStorage || NativeModules.AsyncLocalStorage;
+}
 
 export default RCTAsyncStorage;


src/RCTAsyncStorage.expo.js
```diff
@@ -7,7 +7,18 @@
...
 
-export default RCTAsyncStorage;
\ No newline at end of file
+let RCTAsyncStorage;
+
+// TurboModuleRegistry falls back to NativeModules
+// so we don't have to try go assign NativeModules'
+// counterparts if TurboModuleRegistry would resolve
+// with undefined.
+if (TurboModuleRegistry) {
```
Krizzu, line 18: Looks like `TurboModuleRegistry` [is available to import since 0.60](https://github.com/facebook/react-native/blob/0.60-stable/Libraries/react-native/react-native-implementation.js#L292). Wouldn't it break anything on Expo end?

sjchmiela: I couldn't think of any scenario when this would break:
- if a project's RN is below 0.60 (in Expo terms that's SDKs `(…, 35]`)
  - most probably the project wouldn't use this library, but if it would, `TurboModuleRegistry` would be `undefined` and we'd keep the old behavior of using only `NativeModules` — no change
- if a project's RN is above 0.60 (SDKs `[36, …)`)
  - `TurboModuleRegistry` resolves to a truthy value, we get the new behavior
  - on SDKs where TurboModules aren't enabled (`[36, 38]`)
    - native `AsyncLocalStorage` module is present in `NativeModules`, so `TurboModuleRegistry.get` resolves with this native module (see code for [0.60](https://github.com/facebook/react-native/blob/14c03298d851b1b36854eda994ff62a8406bff56/Libraries/TurboModule/TurboModuleRegistry.js#L20-L25), [0.61](https://github.com/facebook/react-native/blob/86cad7d69099baaaeae10ac7c0912a023101f257/Libraries/TurboModule/TurboModuleRegistry.js#L19-L27), [0.62](https://github.com/facebook/react-native/blob/b9944e54ae35c2beed0e78ea454d871e0fe92ec6/Libraries/TurboModule/TurboModuleRegistry.js#L19-L27) [the `RN$Bridgeless` is never set in Expo])
      ```ts
      export function get<T: TurboModule>(name: string): ?T {
        // Bridgeless mode requires TurboModules
        if (!global.RN$Bridgeless) {
          // Backward compatibility layer during migration.
          const legacyModule = NativeModules[name];
          if (legacyModule != null) {
           return ((legacyModule: any): T);
          }
        }
        
        // ...
      }
      ```
  - on SDKs where TurboModules are enabled (`[39, …)`)
    - native `AsyncLocalStorage` module is no longer present in `NativeModules`, since RN rightfully recognizes it as TurboModule and doesn't initialize it (see [code](https://github.com/facebook/react-native/blob/b9944e54ae35c2beed0e78ea454d871e0fe92ec6/React/CxxBridge/RCTCxxBridge.mm#L741-L744), stayed the same since 0.60)
      ```objc
          if (RCTTurboModuleEnabled() && [moduleClass conformsToProtocol:@protocol(RCTTurboModule)]) {
            continue;
          }
      ```
    - instead, `AsyncLocalStorage` is now available via `turboModuleProxy`
    - `TurboModuleRegistry.get` fails to find the module in `NativeModules` and uses `turboModuleProxy`
    - so when we do `TurboModuleRegistry.get(…)` we'll get the `AsyncLocalStorage` TurboModule

When it comes to testing I could think of two scenarios:
- `RNTester`
  1. Install this library in `RNTester`
  2. Remove library's native code from the iOS project (manually unlink it, mimics Expo environment)
  3. Copy this `RCTAsyncStorage.expo.js` to `RCTAsyncStorage.js`
  4. Verify if using this library still works.
- Expo
  1. Clone branch from this PR https://github.com/expo/expo/pull/9687
  2. Run `yarn` in the project root
  3. Open `ios/Exponent.xcworkspace` and run the app
  4. Go to `apps/native-component-list`, run `expo ios` (given you have `expo-cli` installed, it isn't a workspace dependency)
  5. Notice that `FileSystem` screen (one of the screens that uses `async-storage`) is unavailable (it's greyed out at the bottom of the <kbd>APIs</kbd> tab). You can also open `node_modules/@react-native-community/async-storage/src/RCTAsyncStorage.expo.js` and hit <kbd>save</kbd> for Fast Refresh to kick in — this will show the `Native module is null` error.
  6. Edit `node_modules/@react-native-community/async-storage/src/RCTAsyncStorage.expo.js` so that it contains the code "after" this PR, save the file. Notice the error go away and that `FileSystem` screen is back available.

Krizzu: @sjchmiela Thanks for comprehensive explanation. Just tested this one out on RNTester and everything works.

sjchmiela:
Thank you for merging it and verifying that the fix is correct!

I'm sorry to ask you more of your time, but would you care to publish a new version of `async-storage` with this patch included? At Expo we're preparing for release process and we want to both enable TurboModules and include `async-storage` as a supported library and for that we need to have, well, a release to users point to. 🙂 🙏

Krizzu:
v1.12.0 is out :)

jfrolich:
would it make sense to not put this under an `expo` switch and enable this for non-expo (or ejected) expo apps as well?

Issue #422, opened 2020-08-20
attempt to write a readonly database (code 1032 SQLITE_READONLY_DBMOVED[1032])


kevinranks:
I have users receiving this error: 

`attempt to write a readonly database (code 1032 SQLITE_READONLY_DBMOVED[1032])`

Some Googling and #178 even mentions that this could be related to the default max storage size (6MB). However, I've already implemented this (we're on `v1.7.0` prior to `AsyncStorage_db_size_in_MB` implementation) into our `MainApplication.java` but it has had no affect:

```
@Override
public void onCreate() {
	super.onCreate();
	SoLoader.init(this, /* native exopackage */ false);
	long size = 50L * 1024L * 1024L; // 50 MB
	com.facebook.react.modules.storage.ReactDatabaseSupplier.getInstance(getApplicationContext()).setMaximumSize(size);
}
```

I verified that the size increase was working by implementing a custom NativeModule which calls `ReactDatabaseSupplier` then `getMaximumSize()` to retrieve the size and send it across the bridge to JavaScript. When inside my app and calling `getMaxSize()` it successfully returns 50 MB:

```
@ReactMethod
public void getMaxSize(Callback successCallback) {
	Long maxDatabaseSize = com.facebook.react.modules.storage.ReactDatabaseSupplier.getInstance(reactContext).get().getMaximumSize();
	successCallback.invoke(maxDatabaseSize.doubleValue());
}
```

Does this look correct for verifying that size? Is this error unrelated to Android AsyncStorage size? Looking through the Sentry breadcrumbs this error always seems to occur after the user backgrounds & foregrounds the app following by low memory:

![image](https://user-images.githubusercontent.com/6843831/90810127-b924c300-e2df-11ea-9925-bf8000b3b99b.png)

Krizzu:
Hey @kevinranks 

I'm trying to address this in #423. I'm not able to reproduce this issue myself, so small research with similar problems points it to locale change/db connection not closed properly. Any help here is appreciated 🙏

kevinranks:
@Krizzu We're also unable to reproduce the error despite a lot of trying. If your PR gets merged in we will definitely try it.

baltagih2:
Same issue here

Error
attempt to write a readonly database (code 1032 SQLITE_READONLY_DBMOVED[1032])

```
node_modules/@react-native-community/async-storage/lib/AsyncStorage.js in convertError at line 413:19
function convertError(error): ?Error {
  if (!error) {
[...]

Issue #178, opened 2019-08-02
Failed to open database RKStorage (language)


Bruno-Furtado:
## Current behavior

After changing the device language the app crashs.

## Expected behavior

No impact on the app when user changes the device language.

## Repro steps

1. Install a release build using some language, like `en_US`
2. Kill the app
3. Change the device language, for example `pt_BR`
4. Open the app
4. App crashs

## Environment

- **Async Storage version**: 1.6.1
- **React-Native version**: 0.60.4
- **Platform tested**: Android
- **Logs/Error that are relevant**:

```log
Failed to open database "/data/user/0/com.app/databases/RKStorage".
android.database.sqlite.SQLiteException: Failed to change locale for db "/data/user/0/com.app/databases/RKStorage" to "en_US".
	at android.database.sqlite.SQLiteConnection.setLocaleFromConfiguration(SQLiteConnection.java:394)
	at android.database.sqlite.SQLiteConnection.open(SQLiteConnection.java:219)
	at android.database.sqlite.SQLiteConnection.open(SQLiteConnection.java:194)
	at android.database.sqlite.SQLiteConnectionPool.openConnectionLocked(SQLiteConnectionPool.java:493)
	at android.database.sqlite.SQLiteConnectionPool.open(SQLiteConnectionPool.java:200)
	at android.database.sqlite.SQLiteConnectionPool.open(SQLiteConnectionPool.java:192)
	at android.database.sqlite.SQLiteDatabase.openInner(SQLiteDatabase.java:864)
	at android.database.sqlite.SQLiteDatabase.open(SQLiteDatabase.java:849)
	at android.database.sqlite.SQLiteDatabase.openDatabase(SQLiteDatabase.java:724)
	at android.database.sqlite.SQLiteDatabase.openDatabase(SQLiteDatabase.java:714)
	at android.database.sqlite.SQLiteOpenHelper.getDatabaseLocked(SQLiteOpenHelper.java:295)
	at android.database.sqlite.SQLiteOpenHelper.getWritableDatabase(SQLiteOpenHelper.java:238)
	at com.facebook.react.modules.storage.ReactDatabaseSupplier.ensureDatabase(ReactDatabaseSupplier.java:88)
	at com.facebook.react.modules.storage.AsyncStorageModule.ensureDatabase(AsyncStorageModule.java:432)
	at com.facebook.react.modules.storage.AsyncStorageModule.access$000(AsyncStorageModule.java:38)
[...]

PR #423, opened 2020-08-21
[Android]: Database locale issues

Krizzu:
Summary:
---------

Trying to address: #422 #178 

Those issues we've seen are related to locale changing and database connection not being closed properly. I could not reproduce it myself, so those are just **preventive measures**. Can't tell if this would help for sure.

The fix is to close db connection when activity is destroyed - it'll be reestablished on next startup (in case of locale change, activity is recreated).

~~Another one is to disable [localized collators for db](https://developer.android.com/reference/android/database/sqlite/SQLiteDatabase#NO_LOCALIZED_COLLATORS), which is the most suggested fix for `Failed to change locale for db` error.
I've moved this fix behind a flag that you need to set in `gradle.properties` in order to work.~~

This has been dropped due to possibility of unexpected behavior. [More info about being consistent with locale collation in android docs.](https://developer.android.com/reference/kotlin/android/database/sqlite/SQLiteDatabase#no_localized_collators)

---

If anyone is able to constantly reproduce the issue, please do let me know. You can try out this branch before it's released by pointing your `package.json` dependency to:

```json
"@react-native-community/async-storage": "git+https://github.com/react-native-community/async-storage.git#fix/ensure-db-closed"
```

### DIFF
--- a/android/build.gradle
+++ b/android/build.gradle
@@ -33,6 +33,10 @@ def useDedicatedExecutor = rootProject.hasProperty('AsyncStorage_dedicatedExecut
         ? rootProject.properties['AsyncStorage_dedicatedExecutor']
         : false
 
+def disableLocalizedCollators = rootProject.hasProperty('AsyncStorage_noLocalizedCollators')
+        ? rootProject.properties['AsyncStorage_noLocalizedCollators']
+        : false
+
 apply plugin: 'com.android.library'
 
 android {
@@ -43,6 +47,7 @@ android {
         targetSdkVersion safeExtGet('targetSdkVersion', 28)
         buildConfigField "Long", "AsyncStorage_db_size", "${dbSizeInMB}L"
         buildConfigField("boolean", "AsyncStorage_useDedicatedExecutor", "${useDedicatedExecutor}")
+        buildConfigField("boolean", "AsyncStorage_noLocalizedCollators", "${disableLocalizedCollators}")
     }
     lintOptions {
         abortOnError false
--- a/android/build.gradle
+++ b/android/build.gradle
@@ -33,10 +33,6 @@ def useDedicatedExecutor = rootProject.hasProperty('AsyncStorage_dedicatedExecut
         ? rootProject.properties['AsyncStorage_dedicatedExecutor']
         : false
 
-def disableLocalizedCollators = rootProject.hasProperty('AsyncStorage_noLocalizedCollators')
-        ? rootProject.properties['AsyncStorage_noLocalizedCollators']
-        : false
-
 apply plugin: 'com.android.library'
 
 android {
@@ -47,7 +43,6 @@ android {
         targetSdkVersion safeExtGet('targetSdkVersion', 28)
         buildConfigField "Long", "AsyncStorage_db_size", "${dbSizeInMB}L"
         buildConfigField("boolean", "AsyncStorage_useDedicatedExecutor", "${useDedicatedExecutor}")
-        buildConfigField("boolean", "AsyncStorage_noLocalizedCollators", "${disableLocalizedCollators}")
     }
     lintOptions {
         abortOnError false
--- a/android/src/main/java/com/reactnativecommunity/asyncstorage/ReactDatabaseSupplier.java
+++ b/android/src/main/java/com/reactnativecommunity/asyncstorage/ReactDatabaseSupplier.java
@@ -7,15 +7,14 @@
 
 package com.reactnativecommunity.asyncstorage;
 
-import javax.annotation.Nullable;
-
 import android.content.Context;
 import android.database.sqlite.SQLiteDatabase;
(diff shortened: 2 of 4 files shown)


kevinranks:
Would love to see this reviewed & merged soon if possible. We have many frustrated users that are receiving the issue in #422.

android/src/main/java/com/reactnativecommunity/asyncstorage/ReactDatabaseSupplier.java
```diff
@@ -85,7 +86,13 @@ public void onUpgrade(SQLiteDatabase db, int oldVersion, int newVersion) {
         if (tries > 0) {
           deleteDatabase();
         }
-        mDb = getWritableDatabase();
+
+        if(disableLocalizedCollators) {
```
tido64, line 90: Do we really need the class member?
```suggestion
        if (BuildConfig.AsyncStorage_noLocalizedCollators) {
```

android/build.gradle
```diff
@@ -33,6 +33,10 @@ def useDedicatedExecutor = rootProject.hasProperty('AsyncStorage_dedicatedExecut
         ? rootProject.properties['AsyncStorage_dedicatedExecutor']
         : false
 
+def disableLocalizedCollators = rootProject.hasProperty('AsyncStorage_noLocalizedCollators')
+        ? rootProject.properties['AsyncStorage_noLocalizedCollators']
+        : false
```
tido64, line 38: Nit: We don't really need the ternary but I see we've done the same above...
```suggestion
def disableLocalizedCollators = rootProject.hasProperty('AsyncStorage_noLocalizedCollators')
        && rootProject.properties['AsyncStorage_noLocalizedCollators']
```

tido64 (approved):

android/src/main/java/com/reactnativecommunity/asyncstorage/ReactDatabaseSupplier.java
Krizzu: True, nice catch

android/build.gradle
Krizzu: Yes, I kept it for consistency

Krizzu:
Thanks @tido64 for having a look 🙏 

According to [Android docs, one needs to be consistent with what collation rules were used when creating the database](https://developer.android.com/reference/kotlin/android/database/sqlite/SQLiteDatabase#no_localized_collators). That means, setting this flag to `NO_COLLATION` during runtime could lead to some unexpected behaviors/errors. Besides, the main issue in [this issue](https://github.com/react-native-community/async-storage/issues/178) is `database is locked` error - which is also addressed by this PR.

android/src/main/java/com/reactnativecommunity/asyncstorage/ReactDatabaseSupplier.java
```diff
@@ -7,15 +7,14 @@
...
-import javax.annotation.Nullable;
-
 import android.content.Context;
 import android.database.sqlite.SQLiteDatabase;
 import android.database.sqlite.SQLiteException;
 import android.database.sqlite.SQLiteOpenHelper;
-
 import com.facebook.common.logging.FLog;
 import com.facebook.react.common.ReactConstants;
+import java.io.File;
```
tido64, line 16: Unused?

tido64 (approved):

tido64:
> According to [Android docs, one needs to be consistent with what collation rules were used when creating the database](https://developer.android.com/reference/kotlin/android/database/sqlite/SQLiteDatabase#no_localized_collators). That means, setting this flag to `NO_COLLATION` during runtime could lead to some unexpected behaviors/errors. Besides, the main issue in [this issue](https://github.com/react-native-community/async-storage/issues/178) is `database is locked` error - which is also addressed by this PR.

Does this mean that the problem is that we tried to open a database that was already opened in a different locale? Wouldn't restarting the app already close the database? Or is the implementation such that it is possible to leak connections?

Krizzu:
> Does this mean that the problem is that we tried to open a database that was already opened in a different locale? 

I believe this is the case we see here.

> Wouldn't restarting the app already close the database? Or is the implementation such that it is possible to leak connections?

This is exactly what I think is happening - changing locale restarts the activity, not the app process, so it leaks the db connection.

tido64:
> This is exactly what I think is happening - changing locale restarts the activity, not the app process, so it leaks the db connection.

The repro steps in #178 mentions killing the app though. I thought that would close the connection?

Krizzu:
> The repro steps in #178 mentions killing the app though. I thought that would close the connection?

This is the most confusing part of the whole issue. I also thought the connection will be closed, but apparently it's not 🤷‍♂️ Dunno if we're not missing something here.

Krizzu:
@tido64 I'll go ahead and merge this. Thanks for reviewing 🙏

PR #428, opened 2020-08-27
Real async for web

ferco0:
Summary:
---------
To follow the purpose of the module, this change is necessary due to synchronous nature of localStorage. Uses the localForage module under the hood to abstract methods and technology available. Without break changes.

Test Plan:
----------
Current tests methods cover the changes


Krizzu:
I understand the reason behind this PR, but those changes are too radical. It'll affect current user with changed persistence layer. 

Thanks for your contribution, but I won't merge this. We need to discuss this to find a better way. [(There's a point in Contribution guide about opening issue first, before working on a feature)](https://github.com/react-native-community/async-storage/blob/master/CONTRIBUTING.md#sending-a-pull-request)

ferco0:
Not even a fallback? localstorage has many disadvantages, security it's the most worrisome.

PR #429, opened 2020-08-27
Add flow annotation to index & hooks

tcirstea:
Summary:
---------
Including `react-native-community/async-storage` in a new project with flow doesn't actually work due to the index file not having the flow annotation. (Blank project with `[declarations] node_modules/**` in the flowconfig, flow version 0.132.0)

This change adds flow annotations. Runtime wise this is only a comment and thus can't break anything other than flow itself (which still works).

Screen Shot 2020-08-26 at 6 27 14 PM

Test Plan:
----------
`yarn test` still works.
I'm not completely familiar with how the build process works, but my understanding is that the .flow files in lib will be copied over as is. Thus, the output would look like this, and flow typing would now work:
Screen Shot 2020-08-27 at 11 36 37 AM

If this looks good, I can also bump the version before merging if you'd like.

### DIFF
--- a/src/hooks.js
+++ b/src/hooks.js
@@ -1,5 +1,6 @@
 /**
  * @format
+ * @flow
  */
 
 import AsyncStorage from './AsyncStorage';
--- a/src/index.js
+++ b/src/index.js
@@ -1,5 +1,6 @@
 /**
  * @format
+ * @flow
  */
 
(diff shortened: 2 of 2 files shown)


Krizzu (approved):
Coolio, thanks 👍

PR #446, opened 2020-09-26
Update Installation.md

MohamadKh75:
## Summary:

Hey

Due to https://github.com/facebook/react-native-website/pull/2115, I add a note to the installation docs.

## Test Plan:

`null`

### DIFF
--- a/website/docs/Installation.md
+++ b/website/docs/Installation.md
@@ -32,3 +32,7 @@ react-native link @react-native-community/async-storage
 ```
 
 *Note:* For `Windows` the [manual linking](Linking.md) is currently the only linking option.
+
+### Notes
+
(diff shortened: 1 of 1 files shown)


Krizzu:
Hey @MohamadKh75 ,

What's the reason behind adding this note here? `AsyncStorage` does not use external storage.

MohamadKh75:
Hey @Krizzu 
As you can see in [this discussion](https://github.com/facebook/react-native-website/pull/2115), i just guess it would be good to mention this in installation docs.

Since maybe some users are confused about the storage functionality.

If you guess it's not needed, we can close this.

Krizzu:
Thanks, I guess adding this would make it more confusing than not mentioning permissions at all.

PR #450, opened 2020-09-29
Improved check for invalid parameters

ifsnow:
Summary:
---------

Printing a warning message from `checkValidInput` doesn't seem to work as intended.

In the following case, when an incorrect key or value is passed, `RCTAsyncStorage.multiSet` native function is called and an error occurs unconditionally. This is an unnecessary call.
```js
setItem: function(...) {
    checkValidInput(key, value);
    RCTAsyncStorage.multiSet([[key, value]], function(errors) {
    ...
  }
}
```

I think it's better to throw an error in all cases where parameter check is necessary. This is because the thrown error is displayed in RN's `LogBox` and the logic is not executed after the error.

Test Plan:
----------

None

### DIFF
--- a/src/AsyncStorage.native.js
+++ b/src/AsyncStorage.native.js
@@ -51,7 +51,7 @@ function checkValidInput(usedKey: string, value: any) {
   const isValuePassed = arguments.length > 1;
 
   if (typeof usedKey !== 'string') {
-    console.warn(
+    throw new Error(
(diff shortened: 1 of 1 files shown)


Krizzu:
Hey @ifsnow 

Thanks for raising this up. We choose to show warning instead, as a pointer for dev to fix previous and upcoming issues with values - throwing an error would be to drastic, if you'd upgrade AsyncStorage in project. 

Even better way would be to just stringify/parse data on write/read instead, so one don't even have to remember to do so. But that's something I want address in future releases.

Let's not introduce any new behavior for now, as `AsyncStorage` is going to be moved in its own organization - let's get back to it after it's done.

PR #463, opened 2020-10-16
resolved missing Deploy issue

chiaramooney:
Summary:
---------

Fixes #6239

Resolving error message "ReactNativeAsyncStorage.vcxproj : error MSB4057: The target "Deploy" does not exist in the project" by adding target with Deploy name. 

Test Plan:
----------

To test, add module to React Native app with Windows support. Run app in Windows emulator. Change is successful if app loads in Windows emulator without errors.

PR #467, opened 2020-10-22
chore: Bump RN to 0.62

tido64:
Summary:
---------

Bump the example app to 0.62.

We only bump to 0.62 because react-native-macos 0.63 isn't out yet. To simplify our lives, we stay in sync to avoid any potential issues from having both 0.62 and 0.63 installed.

Test Plan:
----------

The test app must build and boot on the following platforms:
- [x] Android
- [x] iOS
- [x] macOS
- [ ] ~~Web~~ (see https://github.com/react-native-async-storage/async-storage/pull/467#discussion_r511085272)
- [x] Windows


package.json
```diff
@@ -64,26 +64,27 @@
...
+    "@react-native-community/cli-platform-ios": "^4.5.0",
+    "@react-native-community/eslint-config": "^2.0.0",
     "babel-jest": "^26.5.2",
     "babel-plugin-module-resolver": "3.1.3",
     "detox": "16.7.2",
-    "eslint": "5.1.0",
-    "expo": "36.0.2",
-    "flow-bin": "0.105.2",
+    "eslint": "^6.0.0",
+    "expo": "^38.0.10",
```
tido64, line 75: @Krizzu: Do you know how to test this? I couldn't make it work...

Krizzu: We have `expo start:web` script to run web example - maybe try this?

tido64: Finally got it running. Turns out I had a year old expo-cli installed 😛 

I hit other issues though that I can repro on `master`. When I run `start:web`, I get the following error message:

```
Module not found: Can't resolve '@react-native-async-storage/async-storage'
```

in both `async-storage/example/examples/GetSetClear.js` and `async-storage/example/examples/MergeItem.js`. When I tried changing the import to `../src/index`, I got a little further bit hit this instead:

```
Unhandled Rejection (TypeError): Cannot read property 'getItem' of undefined
GetSet.componentWillMount$
/~/example/examples/GetSetClear.js:29
  26 | };
  27 | 
  28 | async componentWillMount() {
> 29 |   const storedNumber = await AsyncStorage.getItem(STORAGE_KEY);
     |                                           ^
  30 |   if (storedNumber) {
  31 |     this.setState({
  32 |       storedNumber,
```

It seems like the web example got broken at some point. @EvanBacon: Is this error obvious to you?

tido64: @Krizzu: If you don't mind, I think we should fix this separately. I've tested all the other platforms, and they're ready to go.

Krizzu: Let's do it 🙏

example/__tests__/App.js
```diff
@@ -79,31 +79,31 @@ describe('Async Storage mock functionality', () => {
   });
 
   describe('Callback based', () => {
-    it('can read/write data to/from storage', done => {
+    it('can read/write data to/from storage', (done) => {
```
Krizzu, line 82: This seems like it's not our eslint config - could you check it out?

example/examples/GetSetClear.js
```diff
@@ -8,76 +8,62 @@
...
 
-  async componentWillMount() {
-    const storedNumber = await AsyncStorage.getItem(STORAGE_KEY);
-    if (storedNumber) {
-      this.setState({
-        storedNumber,
-      });
-    }
-  }
+export default function GetSet(_: Props) {
```
Krizzu, line 18: Yay, functional components 🎉 
nit: You can remove `Props` and this parameter

example/macos/Podfile
```diff
@@ -0,0 +1,17 @@
+require_relative '../../node_modules/react-native-test-app/macos/test_app.rb'
+
+workspace 'AsyncStorageExample.xcworkspace'
+
+use_test_app! do |target|
+  target.app do
+    pod 'RNCAsyncStorage', :path => "../.."
+
+    # This is usually not needed but autolinking seems to be assuming that
+    # node_modules is in '..'
```
Krizzu, line 10: 👍

example/windows/AsyncStorageExample/AsyncStorageExample.vcxproj
```diff
@@ -163,6 +163,7 @@
       npx --no-install react-native bundle --platform windows --entry-file index.js --bundle-output $(MSBuildThisFileDirectory)/Bundle/index.windows.bundle --assets-dest $(MSBuildThisFileDirectory)/Bundle
     </BundleCommand>
   </PropertyGroup>
+  <Target Name="Deploy"/>
```
Krizzu, line 166: 👍 I see this is the same issue [resolved here](https://github.com/react-native-async-storage/async-storage/pull/463)

jest/async-storage-mock.js
```diff
@@ -34,7 +34,7 @@ const asMock = {
   multiSet: jest.fn(_multiSet),
   multiRemove: jest.fn(_multiRemove),
   multiMerge: jest.fn(_multiMerge),
-  useAsyncStorage: jest.fn(key => {
+  useAsyncStorage: jest.fn((key) => {
```
Krizzu, line 37: This seems like our Eslint did not keep in

example/__tests__/App.js
tido64: Yeah, I bumped `@react-native-community/eslint-config` to ^2.0 which bumps Prettier to ^2.0. This is expected.

image

jest/async-storage-mock.js
tido64: See https://github.com/react-native-async-storage/async-storage/pull/467#discussion_r511402295

package.json
tido64: Filed #472

example/examples/GetSetClear.js
tido64: Got a complaint about `componentWillMount` being deprecated so I converted it. Will remove `Props` parameter.

Krizzu (commented):
Fantastic work 🎉 I'm glad for `react-native-test-app` integration!

Let's address web in follow up PR, like you proposed. I have an idea how to fix it.

Also, I know what happened to the formatting - upgrading to latest community eslint config. That version includes Prettier `2.0.2` dependency, which [included a change that generates a lot of churn in codebases, without an option to disable.](https://github.com/prettier/prettier/pull/3903)

I think it's ugly, but nothing we can here. We can, however, disable adding brackets in arrow functions - could you add another entry to `.prettierrc`? 🙏 

```diff
+  "arrowParens": "avoid"
```

tido64:
> I think it's ugly, but nothing we can here. We can, however, disable adding brackets in arrow functions - could you add another entry to `.prettierrc`? 🙏

I'd rather not have a `.prettierrc` to be honest. I kind of agree with this formatting. Makes it consistent for any number of arguments, e.g.:

```js
const single = (value) => value;
const pair = (x, y) => [x, y];
```

The whole point of Prettier is to avoid these discussions. I think we should use the defaults.

Krizzu:
Let's go with the defaults then, like you suggested.

Krizzu (approved):

PR #470, opened 2020-10-23
chore: Semantic release setup

Krizzu:
Summary:
---------

Setting up semantic releases. Every merge to master, after successful tests, means a new release.


tido64 (approved):
Looks fine to me. Do you know what happened to the Android test? I also thought we had macOS ones as well.

Krizzu:
@tido64 Had to disable it because of some error that prevented us from releases. I'm going to address it soon

PR #474, opened 2020-10-24
chore: e2e Android

Krizzu:
Summary:
---------

Bringing back Android e2e tests. Had to do some in-depth testing for environment, and ended up with different macos versions for e2e android and ios (due to different JDK versions).

Additionally, `pods caching`, to speedup iOS builds and tests.

Test Plan:
----------

Green CI.


.circleci/config.yml
```diff
@@ -26,6 +26,13 @@ default config for macOS: &macos_defaults
   macos:
     xcode: '12.1.0'
 
+
+config for macOS (android): &macos_defaults_android
+  <<: *defaults
+  resource_class: 'medium'
+  macos:
+    xcode: '11.5.0'
```
tido64, line 34: Why can't we use `macos_defaults` for Android as well?

.circleci/config.yml
```diff
@@ -41,67 +48,76 @@ default config for android apk builds: &android_defaults
...
-# brew
-save brew cache: &cache_save_brew
-  name: Saving Brew cache
-  paths:
+cache keys:
+  brew ios: &key_brew_ios cache-brew-ios-v3-{{ arch }}
+  brew android: &key_brew_android cache-brew-android-v3-{{ arch }}
+  yarn: &key_yarn cache-yarn-{{ checksum "package.json" }}-{{ arch }}
+  gradle: &key_gradle cache-gradle-{{ checksum "example/android/gradle/wrapper/gradle-wrapper.properties" }}-{{ checksum "package.json" }}-{{ arch }}
+  pods: &key_pods cache-pods-v1-{{ checksum "example/ios/Podfile" }}-{{ checksum "package.json" }}-{{ arch }}
```
tido64, line 56: We should probably check in `Podfile.lock` and use that for checksum, otherwise we can't guarantee that we'll use the cache at all.
```suggestion
  pods: &key_pods cache-pods-v1-{{ checksum "example/ios/Podfile.lock" }}-{{ arch }}
```

I realize this has always been here so maybe we can fix it in a follow up?

.circleci/config.yml
```diff
@@ -203,35 +215,34 @@ jobs:
...
-            yes | sdkmanager "platform-tools" "tools" 1> /dev/null
-            yes | sdkmanager "platforms;android-28" "system-images;android-21;google_apis;x86" 1> /dev/null
-            yes | sdkmanager "emulator" --channel=3 1> /dev/null
-            yes | sdkmanager "build-tools;28.0.3" 1> /dev/null
-            yes | sdkmanager --licenses 1> /dev/null
+            yes | sdkmanager "platform-tools" "tools" >/dev/null
+            yes | sdkmanager "platforms;android-28" "system-images;android-28;default;x86_64" >/dev/null
+            yes | sdkmanager "emulator" --channel=3 >/dev/null
+            yes | sdkmanager "build-tools;28.0.3" >/dev/null
+            yes | sdkmanager --licenses >/dev/null
```
tido64, line 245: We shouldn't redirect `stderr` in case we need to handle them:
```suggestion
            yes | sdkmanager "platform-tools" "tools" 1> /dev/null
            yes | sdkmanager "platforms;android-28" "system-images;android-28;default;x86_64" 1> /dev/null
            yes | sdkmanager "emulator" --channel=3 1> /dev/null
            yes | sdkmanager "build-tools;28.0.3" 1> /dev/null
            yes | sdkmanager --licenses 1> /dev/null
```

.circleci/config.yml
Krizzu: I had to keep it at this version, due to android emulator not working. Maybe with detox upgrade we can use the same version, but I'll have to check it

.circleci/config.yml
Krizzu: `Podfile.lock` is excluded from git, so we cannot use it. Beside, we'd have to make sure we're using the same Pod version locally and on CI. So instead, I fallback to use combination of package.json and Pod file itself

tido64 (approved):

example/android/build.gradle
```diff
@@ -24,3 +24,41 @@ task fetchDependencies() {
...
+            androidExtension.buildTypes.release.signingConfig = androidExtension.signingConfigs.test
+            androidExtension.testBuildType = 'release'
+
+            androidExtension.sourceSets.androidTest.java.srcDirs += "$rootDir/app/src/androidTest/java"
+
+            project.dependencies {
+                androidTestImplementation('com.wix:detox:+')
+            }
+        }
+    }
```
tido64, line 63: @arazabishov: Do you think we can simplify how devs can customize `react-native-test-app`? Not sure if this is normal as far as Gradle configs go but it seems a bit cumbersome to add `signingConfigs` and Detox specific things, imo.

arazabishov: I think it is fine to leverage the flexibility of Gradle when needed, like in this case. If we were about to parameterize all of the things above, we will find ourselves trying to replicate AGP's API. What we can do is provide a convenient way to tap into androidExtension, so users don't have to search for it themselves.

tido64: If this is normal, then I think it's probably fine. We can add a "recipe" in our wiki and point people in that direction when they need to add stuff.

PR #475, opened 2020-10-24
chore: Migrate Android/iOS example apps to react-native-test-app

tido64:
Summary:
---------

Migrate Android and iOS example apps to `react-native-test-app`.

This PR became a bit bigger than anticipated because the iOS tests were randomly failing. I ended up having to bump Detox to latest version and switch to `jest-circus` as test runner.

See also Detox' [Jest setup guide](https://github.com/wix/detox/blob/master/docs/Guide.Jest.md).

Test Plan:
----------

- [x] Android example app should build and boot
- [x] iOS example app should build and boot
- [x] CI should pass


android/build.gradle
```diff
@@ -1,19 +1,37 @@
+import java.nio.file.Paths
+
+def resolveModulePath(packageName) {
```
Krizzu, line 3: Isn't this addressed in [your other PR?](https://github.com/react-native-async-storage/async-storage/pull/481/files) I think this PR will land first

tido64: Yeah, this was built on top of it. The other PR should go in first, then I'll rebase and we can get this merged.

Krizzu (approved):
Looks fantastic 👏  Let's merge the other PR first, then that one

PR #481, opened 2020-10-31
fix(Android): Find react-native using node module resolution

tido64:
Summary:
---------

AsyncStorage assumes that it is installed in `node_modules` in the parent folder. This may not always be the case.

Test Plan:
----------

Android CI should pass.


Krizzu (approved):
Coolio 💪

PR #485, opened 2020-11-06
chore: Ask bug reporters to provide output of `react-native info`

tido64:
## Summary

Ask bug reporters to provide output of `react-native info` so we can better understand their environment.

## Test Plan

This doesn't affect any code.


Krizzu (approved):
Fantastic 🙏

PR #497, opened 2020-11-24
chore: Script files cleanup

Krizzu:
## Summary

Naming and location of scripts/commands was not good, so did a small cleanup. 

- Moved `.circleci/scripts` content to `scripts` in root (and renamed them to `platform_e2e`)
- Each script contains all necessary functionality for e2e (like bulding apk/ipa, running emulator)

## Test Plan

Green CI.


scripts/ios_e2e.sh
```diff
@@ -39,40 +27,46 @@ run_simulator() {
...
   deviceUUID=${deviceInfo[1]}
 
-  echo "[Detox e2e] Booting up $deviceName (id: $deviceUUID)"
+  echo "[Detox e2e] Booting up $SIMULATOR_NAME (id: $deviceUUID)"
 
   # Booting emulator in headless mode
   eval "open $devDir --args -CurrentDeviceUDID $deviceUUID"
-
-  # Decide if should run headless or not
-  if [ "$runHeadless" -eq 0 ]; then
```
Krizzu, line 63: we never used this one, so removed

scripts/android_e2e.sh
```diff
@@ -1,20 +1,19 @@
 #!/bin/bash
```
tido64, line 1: Would be nice if we could convert these scripts to JS so we're tied to having a Mac.

tido64 (approved):

scripts/android_e2e.sh
Krizzu: Good point, but let's do it as a follow up PR

PR #520, opened 2021-01-09
Fix mergeItem API documentation

aibekmamatov:
## FIX mergeItem method API documentation

Missed property `hobby` on console.log after merging items


Krizzu (approved):
Good catch, thanks!

PR #524, opened 2021-01-16
Migrate Windows test app to `react-native-test-app`

tido64:
## Summary

Migrates Windows test app to `react-native-test-app`.

## Remaining Work

- [x] ~~Figure out why AsyncStorage isn't autolinked~~ Resolution: Requires bumping `react-native-windows` to 0.63
- [x] Wait for Windows template fix: https://github.com/microsoft/react-native-test-app/pull/274

## Test Plan

1. Build Windows example app:
    ```
    yarn
    yarn install-windows-test-app -p example/windows
    start example/windows/ReactTestApp.sln
    yarn start:windows
    ```
2. Make sure the app works as expected

## Screenshots

Since we're also bumping `react-native`, we should also make sure we haven't broken other platforms.

| Android | iOS | macOS | Windows |
|:-:|:-:|:-:|:-:|
| Screenshot_1611676595 | Simulator Screen Shot - iPhone 12 Pro - 2021-01-26 at 16 46 56 | Screenshot 2021-01-26 at 16 52 08 | image |


Krizzu (approved):
Great work 💪

Issue #498, opened 2020-11-27
"Attempt to re-open an already-closed object: SQLiteDatabase" error on a Play Store pre-launch tests


keeprock:
## Current behavior

In a process of releasing a bundle to the Play Store, I received an errors on a pre-launch tests. It shows up as : `«attempt to re-open an already-closed object: SQLiteDatabase /data/user/0/com.app/databases/RKStorage»`

## Expected behavior

App should not get any errors.

## Repro steps

I've got two separate angles at which we can look at this problem.

1. Automated Sentry error response:

My Sentry error system caught an exception, which may happen then an app has been closed after being put into a background (by pressing Home, I guess)

![image](https://user-images.githubusercontent.com/5209820/100455305-8739ae80-30cf-11eb-9532-87167d18f831.png)

At first glance, the app is already to be closed, so why even bother? I don't know what the user actually saw, which makes me worry a bit. It doesn't happen before.

2. Pre-launch tests from the Play Store

Also, there are pre-launch tests, which are mandatory before uploading an app to the Play Store. And they are showing the same errors.

![image](https://user-images.githubusercontent.com/5209820/100455467-cbc54a00-30cf-11eb-97af-6ec573c40866.png)

Related issue - https://github.com/firebase/quickstart-unity/issues/861.

Looks like a problem with pre-release test code on a Google side. But worth looking at any way, cause `async-library` being caught in an error log.

Thanks!

## Environment

- **Platforms tested**:
  - [8.1.0] Android Nokia 1
  - [ ] iOS
  - [ ] macOS
  - [ ] Windows
- **AsyncStorage version**: 1.12.1
- **Environment**:
react-native: =0.63.3 => 0.63.3
- **Logs/Errors that may be relevant**: 
```
An error occurred while executing doInBackground()

android.os.AsyncTask$3 in done at line 353

    java.util.concurrent.FutureTask in finishCompletion at line 383
    java.util.concurrent.FutureTask in setException at line 252
    java.util.concurrent.FutureTask in run at line 271
    com.reactnativecommunity.asyncstorage.AsyncStorageModule$SerialExecutor$1 in run at line 63
    java.util.concurrent.ThreadPoolExecutor in runWorker at line 1162
    java.util.concurrent.ThreadPoolExecutor$Worker in run at line 636
    java.lang.Thread in run at line 764
```
```
IllegalStateException

attempt to re-open an already-closed object: SQLiteDatabase: /data/user/0/com.app/databases/RKStorage
[...]

Issue #355, opened 2020-05-13
The connection pool for database 'xxx' has been unable to grant a connection to thread 376 (pool-5-thread-1) with flags 0x2 for 180.01001 seconds


anthinkingcoder:
## Current behavior
`setItem` not callback 

## Expected behavior
`setItem` callback 

## Repro steps
when excute `mReactDatabaseSupplier.get().setTransactionSuccessful();`,
I found `get()` not enter。it‘s block，maybe not get` lock`。
1. add log
![C7BAAFBD-ED33-499E-9A71-645F17883000](https://user-images.githubusercontent.com/17498633/81768896-c0814700-950e-11ea-8c2c-02f2d3de4d9f.png)
![4415314F-9268-4DD5-BDD6-6422014FF2BA](https://user-images.githubusercontent.com/17498633/81769220-a1cf8000-950f-11ea-876b-9ee03d3b0ae8.png)

run it, show the log。`mReactDatabaseSupplier.get()` not log `get lock AsyncTask #1`，it‘s block。
![84A9EC98-EF01-4DAE-B9EB-BE60C3B1DCFB](https://user-images.githubusercontent.com/17498633/81769448-376b0f80-9510-11ea-9f6d-713bbea9d9ea.png)

2. use `Lock` instead of `synchronized`
get()
![image](https://user-images.githubusercontent.com/17498633/81772267-0393e800-9518-11ea-9722-22e4a9ea172b.png)
ensureDatabase()
![image](https://user-images.githubusercontent.com/17498633/81772300-18707b80-9518-11ea-89e1-1f62ec3a2582.png)

it's ok。
![8F55E933-18AD-431A-933A-D0E803F78BCF](https://user-images.githubusercontent.com/17498633/81772447-6dac8d00-9518-11ea-919e-d0c7cf6a8767.png)

## Environment

- **Async Storage version**: 1.9.0
- **React-Native version**:0.59 
- **Platform tested**: Android9
- **Logs/Error that are relevant**:

Krizzu:
@anthinkingcoder Hey, thanks for the insight. We had similar issue before, hence we created [Dedicated Executor pool for this case](https://github.com/react-native-community/async-storage/blob/master/docs/advanced/DedicatedExecutor.md). Can you give it a try?

anthinkingcoder:
@Krizzu  I tried this method, but I found it was not possible。Only changing the lock mechanism can solve my problem。

PR #528, opened 2021-01-19
feat(android): Use Room library for persistent storage

Krizzu:
## Summary

Current implementation of Sqlite access is cumbersome, error prone and produces [some](https://github.com/react-native-async-storage/async-storage/issues/498) hard to reproduce [errors](https://github.com/react-native-async-storage/async-storage/issues/355) to users. This PR aims to improve our native implementation of SQLite access and thread control. 

### About

This PR introduced the "next" storage implementation. [Room](https://developer.android.com/training/data-storage/room) has been used for database access, and [Coroutines](https://kotlinlang.org/docs/coroutines-overview.html) to handle asynchronous work.

I tried to reuse old database file, but due to it having a SQLite bug (where `primary key` can be null, which is a SQLite violation that Room is guarded against) I had to provide a migration path to copy over values to new database. This operation is done only once, the very first time feature is switched on.

**No developer migration is necessary to start using this feature**. It means that, once switched on, the same data stored so far will be available to use. No API change on JS side is involved. One notable change that can be somehow breaking is how this feature handle AsyncStorage rules violation - if `key` used is not string (or null) OR `value` is not a string, then error with proper message is thrown. No more cryptic sqlite errors.

As part of this PR and feature, some new enhancement:

- Updated the docs, to describe usage of the feature and also added an Android section to Brownfield integration doc, to show how one can access storage from native side.
- added new section in test app, called Basic, to test all basic test for AsyncStorage (get/set/remove, merge, getKeys, clearing db and error validation)
- added native unit tests to validate Storage access (since it can be used from native side as well).

### Test plan

Green CI, manual test on the Example App.

### DIFF
--- a/android/src/main/java/com/reactnativecommunity/asyncstorage/AsyncStorageModule.java
+++ b/android/src/main/java/com/reactnativecommunity/asyncstorage/AsyncStorageModule.java
@@ -46,37 +46,6 @@ public final class AsyncStorageModule
   private ReactDatabaseSupplier mReactDatabaseSupplier;
   private boolean mShuttingDown = false;
 
-  // Adapted from https://android.googlesource.com/platform/frameworks/base.git/+/1488a3a19d4681a41fb45570c15e14d99db1cb66/core/java/android/os/AsyncTask.java#237
-  private class SerialExecutor implements Executor {
-    private final ArrayDeque<Runnable> mTasks = new ArrayDeque<Runnable>();
-    private Runnable mActive;
-    private final Executor executor;
-
-    SerialExecutor(Executor executor) {
-      this.executor = executor;
-    }
-
-    public synchronized void execute(final Runnable r) {
-      mTasks.offer(new Runnable() {
-        public void run() {
-          try {
-            r.run();
-          } finally {
-            scheduleNext();
-          }
-        }
-      });
-      if (mActive == null) {
-        scheduleNext();
-      }
-    }
-    synchronized void scheduleNext() {
-      if ((mActive = mTasks.poll()) != null) {
-        executor.execute(mActive);
-      }
-    }
-  }
-
   private final SerialExecutor executor;
 
   public AsyncStorageModule(ReactApplicationContext reactContext) {
--- a/android/src/main/java/com/reactnativecommunity/asyncstorage/next/ArgumentHelpers.kt
+++ b/android/src/main/java/com/reactnativecommunity/asyncstorage/next/ArgumentHelpers.kt
@@ -1,0 +1,50 @@
+package com.reactnativecommunity.asyncstorage.next
+
+import com.facebook.react.bridge.Arguments
+import com.facebook.react.bridge.ReadableArray
+
+fun ReadableArray.toEntryList(): List<Entry> {
+    val list = mutableListOf<Entry>()
+    for (keyValue in this.toArrayList()) {
+        if (keyValue !is ArrayList<*>) {
+            throw AsyncStorageError.invalidKeyValueFormat()
+        }
+        val key = keyValue[0]
+        val value = keyValue[1]
+
+        if (key == null || key !is String) {
+            throw AsyncStorageError.keyIsNull()
+        }
+
+        if (value !is String) {
+            throw AsyncStorageError.valueNotString(key)
+        }
+
+        list.add(Entry(key, value))
+    }
+    return list
+}
+
+fun ReadableArray.toKeyList(): List<String> {
+    val list = this.toArrayList().toList()
+
+    for (item in list) {
+        if (item !is String) {
+            throw AsyncStorageError.keyIsNull()
+        }
+    }
+    return list as List<String>
+}
+
+fun List<Entry>.toKeyValueArgument(): ReadableArray {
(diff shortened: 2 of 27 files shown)


android/src/main/java/com/reactnativecommunity/asyncstorage/AsyncStorageModule.java
```diff
@@ -46,37 +46,6 @@
   private ReactDatabaseSupplier mReactDatabaseSupplier;
   private boolean mShuttingDown = false;
 
-  // Adapted from https://android.googlesource.com/platform/frameworks/base.git/+/1488a3a19d4681a41fb45570c15e14d99db1cb66/core/java/android/os/AsyncTask.java#237
-  private class SerialExecutor implements Executor {
```
Krizzu, line 50: Detox is using this implementation detail in its environment setup, so I had to make this class public and available in Next storage (even tho it's not used)

tido64: Can you add this comment in the code so it doesn't get lost?

android/src/main/java/com/reactnativecommunity/asyncstorage/next/ArgumentHelpers.kt
```diff
@@ -0,0 +1,86 @@
...
+fun ReadableArray.toEntryList(): List<Entry> {
+    val list = mutableListOf<Entry>()
+    for (keyValue in this.toArrayList()) {
+        if (keyValue !is ArrayList<*> || keyValue.size != 2) {
+            throw AsyncStorageError.invalidKeyValueFormat()
+        }
+        val key = keyValue[0]
+        val value = keyValue[1]
+
+        if (key == null || key !is String) {
```
tido64, line 17: Do we still need the null check here? Won't the second check cover that case?

android/src/main/java/com/reactnativecommunity/asyncstorage/next/ErrorHelpers.kt
```diff
@@ -0,0 +1,39 @@
...
+
+        fun keyNotString() = AsyncStorageError("Provided key is not string. Only strings are supported as storage key.")
+
+        fun valueNotString(key: String?): AsyncStorageError {
+            val detail = if (key == null) "Provided value" else "Value for key \"$key\""
+            return AsyncStorageError("$detail is not a string. Only strings are supported as a value.")
+        }
+
+        fun invalidKeyValueFormat() =
+            AsyncStorageError("Invalid key-value format. Expected a list of [key, value] list.")
```
tido64, line 36: Should these checks be done on the JS side so we don't have to duplicate the code on all the platforms?

android/src/test/java/com/reactnativecommunity/asyncstorage/next/StorageTest.kt
```diff
@@ -0,0 +1,141 @@
...
+import org.json.JSONObject
+import org.junit.After
+import org.junit.Before
+import org.junit.Test
+import org.junit.runner.RunWith
+import kotlin.random.Random
+
+@ExperimentalCoroutinesApi
+@RunWith(AndroidJUnit4::class)
+class AsyncStorageAccessTest {
```
tido64, line 18: Should we also add tests for migration?

example/android/gradle.properties
```diff
@@ -22,6 +22,7 @@
 
 # Enable dedicated thread pool executor
 AsyncStorage_dedicatedExecutor=true
+AsyncStorage_useNextStorage=true
```
tido64, line 25: Should this be flippable so we can test both paths on CI?

tido64 (approved):
LGTM, but I haven't tested it yet. I'll try to find some time to do so but will tentatively approve in the meantime. 😄

android/src/main/java/com/reactnativecommunity/asyncstorage/next/ErrorHelpers.kt
Krizzu: I agree, but that'd be a breaking change, right? Right now, we have warnings on JS side

example/android/gradle.properties
Krizzu: on Example app, by default this will be true (so one can develop next version), while our E2E apk build job will turn it off via command line

android/src/main/java/com/reactnativecommunity/asyncstorage/next/ErrorHelpers.kt
Krizzu: I think we could do this as a follow up PR, where we'll throw errors instead of warning from JS side, so that we force the correct format of data

android/src/test/java/com/reactnativecommunity/asyncstorage/next/StorageTest.kt
Krizzu: I'd say that it could be valuable if we would do some manipulation of the data while migrating. Right now, we simply copy data from older storage to new one as it is

tido64 (approved):
LGTM

PR #544, opened 2021-02-12
chore: remove .expo extension in favor of generic fallback to legacy native module mechanism

brentvatne:
## Summary

Expo is moving away from `.expo.js` files, so for async-storage to continue working I have moved the environment check to a `shouldFallbackToLegacyNativeModule()` function instead of `RCTAsyncStorage.expo.js`. This function is only invoked if the native modules that ship with this library are not defined.

## Test Plan

- `yarn test`
- `yarn start`
- `yarn start:ios`
- Run example in Android Studio
- I don't have a Windows machine so I could not try Windows, but it will work the same as iOS/Android for the code changed and I imagine CI will cover that.


src/shouldFallbackToLegacyNativeModule.js
```diff
@@ -0,0 +1,20 @@
...
+  /**
+   * Expo managed apps don't include the @react-native-async-storage/async-storage
+   * native modules yet, but the API interface is the same, so we can use the version
+   * exported from React Native still.
+   *
+   * If in future releases (eg: @react-native-async-storage/async-storage >= 2.0.0) this
+   * will likely not be valid anymore, and the package will need to be included in the Expo SDK
+   * to continue to work.
+   */
+  if (expoExecutionEnvironment && ['storeClient', 'standalone'].includes(expoExecutionEnvironment)) {
```
Krizzu, line 15: @brentvatne 
Are those execution envs available in all expo versions? Or is there a min version support for it?

brentvatne: only in sdk 40+ -- but a version of async storage that includes this code will only ever run in sdk 41+ apps, because users install a compatible version of native modules for given sdks using `expo install`

Krizzu (approved):
This looks good, thanks 🙏

PR #552, opened 2021-02-22
fix: Expo fallback function exporting

Krizzu:
## Summary

Fixing the export for Expo fallback method, preventing from using Async Storage.
Also, releases should be done after all e2e passed.

## Test Plan

Green CI.


tido64 (approved):
Is CI picking up this breakage now?

brentvatne:
sorry for that, thanks for fixing

brentvatne (approved):

Issue #507, opened 2020-12-18
Why repository was moved out from react-native-community?


FRizzonelli:
Hello guys!

Just a genuine question here. I'd like to know why repository was moved again. Back in the days we're used to install it directly from react-native and, fortunately, RN team decide to move it out from the core which is a super good thing! 
It was moved to @react-native-community organization which I believe is a good place to keep repos and materials and has a lot of collaborators.

I'd like to know what is the reason of this change, especially because I think it'll be complicated and confusing for new developers to keep track of this new change.

Thanks for any kind of response and feel free to close if this isn't the right place to ask! Also thank you for all the effort you put in open source software ❤️

PR #556, opened 2021-03-03
fix: Cannot find module

haruelrovix:


tido64:
Thanks for your contribution but I think you've got this backwards.  AsyncStorage was moved out of react-native-community into its own org. You can see discussion #507 for more details.

PR #559, opened 2021-03-13
Created Android scoped storage migration method

EvanBacon:
## Summary

Expo users who publish an app with `expo build:android` will use a database file like `RKStorage-scoped-experience- <encoded-project-slug>` when they eject, the 
database file will be `RKStorage`, this effectively prevents many users from being able to eject https://github.com/expo/expo/issues/8220 and use this package.

This PR introduces parity for an [iOS feature](https://github.com/react-native-async-storage/async-storage/blob/af0677c43e5087404b0def0dd084a69e1930cbf5/ios/RNCAsyncStorage.m#L395-L399) that migrates the database directory.

1. Check to see if `RKStorage` file exists (if not bail out).
2. Query all files in the database directory starting with `RKStorage-scoped-experience-`
3. Get the most recently modified file to migrate. This is useful because users who change their slug and republish may have multiple database files and they'll want to use the most recent one.
4. Copy the contents of this file to `RKStorage`
5. Delete all of the scoped database files. This creates parity with iOS.

## Test Plan

Testing this feature is a little _involved_.

### Managed -> Bare

0. Delete the app -- effectively deleting all of it's data and the `RKStorage` file.
1. Comment out the line `AsyncStorageMigration.migrate(reactContext);`
2. Change ["RKStorage"](https://github.com/react-native-async-storage/async-storage/blob/87805694cc8ee07c594ef4c8f56caf4163acb498/android/src/main/java/com/reactnativecommunity/asyncstorage/ReactDatabaseSupplier.java#L25) to something like `RKStorage-scoped-experience-foobar-1`
3. Build the app, make a few modifications.
4. Repeat the last two steps with a new storage name (`RKStorage-scoped-experience-foobar-2`) -- This is to test that the most recent slug gets used.
5. Uncomment the line `AsyncStorageMigration.migrate(reactContext);` and change the database name back to `RKStorage`.
6. Build the app, you'll notice the logs print out info about the migration (ScopedStorageMigration log tag). The AsyncStorage in-app should have changed to match the most recent modifications from `RKStorage-scoped-experience-foobar-2`.
7. Reload the app, you should see no longs related to the migration.

### Bare only

0. Delete the app -- effectively deleting all of it's data and the `RKStorage` file.
1. Build the app, you should see a log "No scoped database found" -- This is what users who run `AsyncStorage` for the first time without ever using Expo will see.
2. Reload the app and there should be no logs related to the migration.

### DIFF
--- a/android/src/main/java/com/reactnativecommunity/asyncstorage/AsyncStorageExpoMigration.java
+++ b/android/src/main/java/com/reactnativecommunity/asyncstorage/AsyncStorageExpoMigration.java
@@ -6,8 +6,6 @@ import android.util.Log;
 
 import androidx.annotation.RequiresApi;
 
-import com.facebook.react.modules.storage.ReactDatabaseSupplier;
-
 import java.io.File;
 import java.io.FileInputStream;
 import java.io.FileOutputStream;
--- a/android/src/main/java/com/reactnativecommunity/asyncstorage/AsyncStorageExpoMigration.java
+++ b/android/src/main/java/com/reactnativecommunity/asyncstorage/AsyncStorageExpoMigration.java
@@ -19,7 +19,7 @@ import java.util.ArrayList;
 
 // A utility class that migrates a scoped AsyncStorage database to RKStorage.
 // This utility only runs if the RKStorage file has not been created yet.
-public class AsyncStorageMigration {
+public class AsyncStorageExpoMigration {
     static final String LOG_TAG = "AsyncStorageExpoMigration";
 
     public static void migrate(Context context) {
--- a/android/src/main/java/com/reactnativecommunity/asyncstorage/AsyncStorageMigration.java
+++ b/android/src/main/java/com/reactnativecommunity/asyncstorage/AsyncStorageMigration.java
@@ -1,0 +1,160 @@
+package com.reactnativecommunity.asyncstorage;
+
+import android.content.Context;
+import android.os.Build;
+import android.util.Log;
+
+import androidx.annotation.RequiresApi;
+
+import com.facebook.react.modules.storage.ReactDatabaseSupplier;
+
+import java.io.File;
+import java.io.FileInputStream;
+import java.io.FileOutputStream;
+import java.io.IOException;
+import java.nio.channels.FileChannel;
+import java.nio.file.Files;
+import java.nio.file.attribute.BasicFileAttributes;
+import java.util.ArrayList;
+
+// A utility class that migrates a scoped AsyncStorage database to RKStorage.
+// This utility only runs if the RKStorage file has not been created yet.
+public class AsyncStorageMigration {
+    static final String LOG_TAG = "ScopedStorageMigration";
+
+    private static Context mContext;
+
+    public static void migrate(Context context) {
+        mContext = context;
+
+        // Only migrate if the default async storage file does not exist.
+        if (isAsyncStorageDatabaseCreated()) {
+            return;
+        }
+
+        ArrayList<File> expoDatabases = getExpoDatabases();
+
+        File expoDatabase = getLastModifiedFile(expoDatabases);
+
+        if (expoDatabase == null) {
+            Log.v(LOG_TAG, "No scoped database found");
+            return;
+        }
+
+        try {
+            // Create the storage file
+            ReactDatabaseSupplier.getInstance(mContext).get();
+            copyFile(new FileInputStream(expoDatabase), new FileOutputStream(mContext.getDatabasePath(ReactDatabaseSupplier.DATABASE_NAME)));
+        } catch (Exception e) {
+            Log.v(LOG_TAG, "Failed to move scoped database");
+            e.printStackTrace();
+            return;
+        }
+
+        try {
+            for (File file : expoDatabases) {
+                if (file.delete()) {
+                    Log.v(LOG_TAG, "Deleted scoped database " + file.getName());
+                } else {
+                    Log.v(LOG_TAG, "Failed to delete scoped database " + file.getName());
+                }
+            }
+        } catch (Exception e) {
+            e.printStackTrace();
+        }
+
+        Log.v(LOG_TAG, "Completed the scoped AsyncStorage migration");
+    }
+
+    private static boolean isAsyncStorageDatabaseCreated() {
+        return mContext.getDatabasePath(ReactDatabaseSupplier.DATABASE_NAME).exists();
+    }
+
+    // Find all database files that the user may have created while using Expo.
+    private static ArrayList<File> getExpoDatabases() {
+        ArrayList<File> scopedDatabases = new ArrayList<>();
+        try {
+            File databaseDirectory = mContext.getDatabasePath("noop").getParentFile();
+            File[] directoryListing = databaseDirectory.listFiles();
+            if (directoryListing != null) {
+                for (File child : directoryListing) {
+                    // Find all databases matching the Expo scoped key, and skip any database journals.
+                    if (child.getName().startsWith("RKStorage-scoped-experience-") && !child.getName().endsWith("-journal")) {
+                        scopedDatabases.add(child);
+                    }
+                }
+            }
+        } catch (Exception e) {
+            // Just in case anything happens catch and print, file system rules can tend to be different across vendors.
+            e.printStackTrace();
+        }
+        return scopedDatabases;
+    }
+
+    // Returns the most recently modified file.
+    // If a user publishes an app with Expo, then changes the slug 
+    // and publishes again, a new database will be created. 
+    // We want to select the most recent database and migrate it to RKStorage.
+    private static File getLastModifiedFile(ArrayList<File> files) {
+        if (files.size() == 0) {
+            return null;
+        }
+        long lastMod = -1;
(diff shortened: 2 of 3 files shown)


android/src/main/java/com/reactnativecommunity/asyncstorage/AsyncStorageMigration.java
```diff
@@ -0,0 +1,161 @@
...
+        mContext = context;
+
+        // Only migrate if the default async storage file does not exist.
+        if (isAsyncStorageDatabaseCreated()) {
+            return;
+        }
+
+        ArrayList<File> expoDatabases = getExpoDatabases();
+
+        File expoDatabase = getLastModifiedFile(expoDatabases);
```
brentvatne, line 37: from the PR description:

> Get the most recently modified file to migrate. This is useful because users who change their slug and republish may have multiple database files and they'll want to use the most recent one.

to elaborate on this, as I saw it, we had two options here:

1) embed the "experienceId" ([see here](https://github.com/expo/expo/blob/9c26fbc09d2ec8c7b48d5e7d79bf99631b7b84cc/android/expoview/src/main/java/versioned/host/exp/exponent/modules/internal/ExponentAsyncStorageModule.java#L21-L24) for how it's used in managed Expo apps
2) use something like last modified time as a heuristic for determining which database was used

(2) should be sufficient and allows us to solve this problem in one place - without depending on external scripts to embed the experienceId somewhere in an ejected application, so I suggested that we proceed with this approach

android/src/main/java/com/reactnativecommunity/asyncstorage/AsyncStorageMigration.java
```diff
@@ -0,0 +1,161 @@
...
+        } catch (Exception e) {
+            Log.v(LOG_TAG, "Failed to migrate scoped database " + expoDatabase.getName());
+            e.printStackTrace();
+            return;
+        }
+
+        try {
+            for (File file : expoDatabases) {
+                if (file.delete()) {
+                    Log.v(LOG_TAG, "Deleted scoped database " + file.getName());
```
brentvatne, line 58: let's leave it in place rather than deleting

brentvatne: changed my mind and it should delete, to behave the same as the ios script. thanks for reverting :)

android/src/main/java/com/reactnativecommunity/asyncstorage/AsyncStorageMigration.java
Krizzu: I like the second option more too

android/src/main/java/com/reactnativecommunity/asyncstorage/AsyncStorageMigration.java
```diff
@@ -0,0 +1,161 @@
...
+        File expoDatabase = getLastModifiedFile(expoDatabases);
+
+        if (expoDatabase == null) {
+            Log.v(LOG_TAG, "No scoped database found");
+            return;
+        }
+
+        try {
+            // Create the storage file
+            ReactDatabaseSupplier.getInstance(mContext).get();
```
Krizzu, line 46: Let's use supplier from this library, not RN

android/src/main/java/com/reactnativecommunity/asyncstorage/AsyncStorageMigration.java
```diff
@@ -0,0 +1,161 @@
...
+import java.io.FileOutputStream;
+import java.io.IOException;
+import java.nio.channels.FileChannel;
+import java.nio.file.Files;
+import java.nio.file.attribute.BasicFileAttributes;
+import java.util.ArrayList;
+
+// A utility class that migrates a scoped AsyncStorage database to RKStorage.
+// This utility only runs if the RKStorage file has not been created yet.
+public class AsyncStorageMigration {
```
Krizzu, line 22: Since this migration is Expo related, can we have it reflected in naming?

android/src/main/java/com/reactnativecommunity/asyncstorage/AsyncStorageMigration.java
```diff
@@ -0,0 +1,161 @@
...
+    static final String LOG_TAG = "ScopedStorageMigration";
```
Krizzu, line 23: Let's have a tag that's related to AsyncStorage, for example `AsyncStorageExpoMigration`

android/src/main/java/com/reactnativecommunity/asyncstorage/AsyncStorageMigration.java
```diff
@@ -0,0 +1,161 @@
...
+
+    private static Context mContext;
```
Krizzu, line 25: nitty nit:  
Saving context in local fields is not best practice and should be avoided. It's used in private methods, so you can pass it around from `migrate` argument.

android/src/main/java/com/reactnativecommunity/asyncstorage/AsyncStorageMigration.java
```diff
@@ -0,0 +1,161 @@
...
+            e.printStackTrace();
+            return -1;
+        }
+    }
+
+    @RequiresApi(Build.VERSION_CODES.O)
+    private static long getLastModifiedTimeFromBasicFileAttrs(File file) {
+        try {
+            BasicFileAttributes attr = Files.readAttributes(file.toPath(), BasicFileAttributes.class);
+            return attr.creationTime().toMillis();
```
Krizzu, line 136: I think you can inline those statements to `getLastModifiedTimeInMillis`

android/src/main/java/com/reactnativecommunity/asyncstorage/AsyncStorageMigration.java
```diff
@@ -0,0 +1,161 @@
...
+    // Find all database files that the user may have created while using Expo.
+    private static ArrayList<File> getExpoDatabases() {
+        ArrayList<File> scopedDatabases = new ArrayList<>();
+        try {
+            File databaseDirectory = mContext.getDatabasePath("noop").getParentFile();
+            File[] directoryListing = databaseDirectory.listFiles();
+            if (directoryListing != null) {
+                for (File child : directoryListing) {
+                    // Find all databases matching the Expo scoped key, and skip any database journals.
+                    if (child.getName().startsWith("RKStorage-scoped-experience-") && !child.getName().endsWith("-journal")) {
```
Krizzu, line 83: nice catch with journal file 👍

Krizzu (commented):
Nice one Evan 👍 I just left few comments to make it bit more explicit that this migration is related to Expo.

Krizzu (commented):
Looks good 👍 
@tido64 mind having a look as well?

android/src/main/java/com/reactnativecommunity/asyncstorage/AsyncStorageExpoMigration.java
```diff
@@ -0,0 +1,154 @@
...
+
+    private static boolean isAsyncStorageDatabaseCreated(Context context) {
+        return context.getDatabasePath(ReactDatabaseSupplier.DATABASE_NAME).exists();
+    }
+
+    // Find all database files that the user may have created while using Expo.
+    private static ArrayList<File> getExpoDatabases(Context context) {
+        ArrayList<File> scopedDatabases = new ArrayList<>();
+        try {
+            File databaseDirectory = context.getDatabasePath("noop").getParentFile();
```
tido64, line 72: Are we passing in `noop` because it doesn't matter and we just want the parent path? I.e. Android doesn't actually check whether a database named "noop" exists?

android/src/main/java/com/reactnativecommunity/asyncstorage/AsyncStorageExpoMigration.java
```diff
@@ -0,0 +1,154 @@
...
+        }
+        return scopedDatabases;
+    }
+
+    // Returns the most recently modified file.
+    // If a user publishes an app with Expo, then changes the slug 
+    // and publishes again, a new database will be created. 
+    // We want to select the most recent database and migrate it to RKStorage.
+    private static File getLastModifiedFile(ArrayList<File> files) {
+        if (files.size() == 0) {
```
tido64, line 94: Would it make sense to add a special case for `files.size() == 1`?

tido64 (approved):
Looks fine to me. Just a couple of questions to educate myself 😄

android/src/main/java/com/reactnativecommunity/asyncstorage/AsyncStorageExpoMigration.java
EvanBacon: correct

android/src/main/java/com/reactnativecommunity/asyncstorage/AsyncStorageExpoMigration.java
EvanBacon: doesn't seem necessary, even the case for zero isn't strictly required.

brentvatne:
thanks a lot for reviewing, @Krizzu and @tido64!

PR #573, opened 2021-03-28
feat: add exception when array is passed for 2nd arg

### LINKED ISSUE #519: Gracefully handle the case where inputs to multiSet are not of the type Array<Array<String>>
## Motivation
At present, React Native applications will crash when `AsyncStorage.multiSet` is called with invalid arguments, for example,  `AsyncStorage.multiSet(['a', 'b'], ['c', 'd'])`. Note that in this case, there are 2 arguments provided to the function for the `keyValuePairs` required to store as data. However, this happens to be a typo and the developer actually intended to wrap it with square brackets, but failed to do so. 

Since there is no error thrown, the application literally just crashes and makes debugging almost impossible.

## Description
An exception should be raised when the type of the inputs to the keyValuePairs argument of multiSet is incorrect, instead of simply letting the application crash.

## New feature implementation
Type comparison of the [first argument](https://github.com/react-native-async-storage/async-storage/blob/master/src/AsyncStorage.js#L135-L138) would be crucial in correcting this.

kiranjd:
- For cases when two or more key values pairs are passed unintentionally as arguements to multiSet.

## Summary
resolves #519

## Test Plan
As I ran and went through the e2e tests, observed that there are no tests aded for `multiSet` and `multiGet`. As this is an exception that is being added to `multiSet`, would not be ideal to create the tests for the same.

### DIFF
--- a/src/AsyncStorage.native.js
+++ b/src/AsyncStorage.native.js
@@ -69,6 +69,14 @@ function checkValidInput(usedKey: string, value: any) {
   }
 }
 
+function checkValidArgs(args: Array<Array<string>>) {
+  if (Array.isArray(args[1])) {
+    throw new Error(
+      '[AsyncStorage] Did you mean to pass an array of key value pairs instead? The second arguement to multiSet cannot be an array\n',
+    );
+  }
+}
+
 /**
  * `AsyncStorage` is a simple, unencrypted, asynchronous, persistent, key-value
  * storage system that is global to the app.  It should be used instead of
@@ -322,6 +330,7 @@ const AsyncStorage = {
     callback?: ?(errors: ?$ReadOnlyArray<?Error>) => void,
   ): Promise<null> {
     return new Promise((resolve, reject) => {
+      checkValidArgs(arguments);
       keyValuePairs.forEach(([key, value]) => {
         checkValidInput(key, value);
       });
--- a/src/AsyncStorage.native.js
+++ b/src/AsyncStorage.native.js
@@ -69,10 +69,26 @@ function checkValidInput(usedKey: string, value: any) {
   }
 }
 
-function checkValidArgs(args: Array<Array<string>>) {
-  if (Array.isArray(args[1])) {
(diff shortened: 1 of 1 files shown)


src/AsyncStorage.native.js
```diff
@@ -322,6 +330,7 @@ const AsyncStorage = {
     callback?: ?(errors: ?$ReadOnlyArray<?Error>) => void,
   ): Promise<null> {
     return new Promise((resolve, reject) => {
+      checkValidArgs(arguments);
```
tido64, line 333: We should check the arguments before creating the promise. Can we move this up a line?

tido64, line 333: I think it's unintuitive to check `arguments`. Can we instead explicitly check `keyValuePairs` and `callback`? For instance:

```js
if (!Array.isArray(keyValuePairs) ||
    keyValuePairs.length === 0 ||
    !Array.isArray(keyValuePairs[0]) {
  throw new Error("Expected key-value pairs as first argument to multiSet");
} else if (callback && typeof callback !== "function") {
  throw new Error("Expected function as second argument to multiSet");
}
```

tido64 (changes requested):

src/AsyncStorage.native.js
kiranjd: That's right about using `arguments. As for the errors, the issue I'm resolving is for cases when the array is not wrapped unintentionally. 
Although, the above error messages would cover that case, won't it be useful to have a `Did you mean to pass an array of key-value pairs instead?` message?

tido64: Yeah, I'm all for adding more messages if it helps the author.

kiranjd: Awesome! have made the changes. Please have a look 🙂

tido64 (approved):
LGTM 🚀  Thanks for making all our users' lives easier.

kiranjd:
Sorry, didn't notice it was approved and not merged. Plase merge and close this PR whenever 🙂

PR #575, opened 2021-04-02
Added testresults.gradle in bundle as it is required to install on android using next storage

gabrielbull:
Install fails on android using next version because of [this line in build.gradle](https://github.com/react-native-async-storage/async-storage/blame/75c571b7b1265e7d5599e0b9e38e19acb6004181/android/build.gradle#L72) created by @Krizzu. It requires a file that is not included in the bundled async storage library.

```
FAILURE: Build completed with 2 failures.

1: Task failed with an exception.
-----------
* Where:
Build file '~/project/node_modules/@react-native-async-storage/async-storage/android/build.gradle' line: 72

* What went wrong:
A problem occurred evaluating project ':react-native-async-storage_async-storage'.
> Could not read script '~/project/node_modules/@react-native-async-storage/async-storage/android/testresults.gradle' as it does not exist.

* Try:
Run with --stacktrace option to get the stack trace. Run with --info or --debug option to get more log output. Run with --scan to get full insights.
==============================================================================
```

image

### DIFF
--- a/package.json
+++ b/package.json
@@ -13,6 +13,7 @@
     "types/",
     "android/src",
     "android/build.gradle",
+    "android/testresults.gradle",
     "ios/",
     "macos/",
     "windows/",


Krizzu (approved):
Nice one, thanks a lot!

package.json
```diff
@@ -13,6 +13,7 @@
     "types/",
     "android/src",
     "android/build.gradle",
+    "android/testresults.gradle",
```
tido64, line 16: @Krizzu: Why is this published? Shouldn't the script just not apply this file?

Krizzu: Yeah, that'd be a follow up PR to apply this conditionally

Krizzu: Or maybe we could just inline it

PR #582, opened 2021-04-09
fix(next storage): Force SQLite checkpoint on Pie and up

Krizzu:
## Summary

From Pie and up, Android started to use Write-ahead logging (WAL) by default, instead of journal rollback. WAL does not write directly to the database file, rather to the supporting WAL file. The content of WAL file is "flushed" to original DB file (this action is called a **checkpoint**), when page size threshold is reached (default is 1000).

Because of that, migration to Next storage might not be successful, because RKStorage database content might still be in WAL file (and Next AsyncStorage uses RKStorage file to rehydrate itself). This PR introduces a helper that will make sure we force **checkpoint**, to always have the latest content on db file, before migrating.

@EvanBacon I suspect we'd want the same thing happen for Expo migration as well?

### DIFF
--- a/android/src/main/java/com/reactnativecommunity/asyncstorage/AsyncLocalStorageUtil.java
+++ b/android/src/main/java/com/reactnativecommunity/asyncstorage/AsyncLocalStorageUtil.java
@@ -155,15 +155,17 @@ public class AsyncLocalStorageUtil {
    * a "checkpoint" and is done automatically (by default) when the WAL file reaches a threshold
    * size of 1000 pages.
    * More here: https://sqlite.org/wal.html
-   * <p>
+   *
    * This helper will force checkpoint on RKStorage, if Next storage file does not exists yet.
    */
   public static void verifyAndForceSqliteCheckpoint(Context ctx) {
+    if(Build.VERSION.SDK_INT < Build.VERSION_CODES.P) {
+        Log.i("AsyncStorage_Next", "SQLite checkpoint not required on this API version.");
+    }
+
     File nextStorageFile = ctx.getDatabasePath("AsyncStorage");
     File currentStorageFile = ctx.getDatabasePath(ReactDatabaseSupplier.DATABASE_NAME);
-    boolean isCheckpointRequired = Build.VERSION.SDK_INT >= Build.VERSION_CODES.P
-            && !nextStorageFile.exists()
-            && currentStorageFile.exists();
+    boolean isCheckpointRequired = !nextStorageFile.exists() && currentStorageFile.exists();
     if (!isCheckpointRequired) {
       Log.i("AsyncStorage_Next", "SQLite checkpoint not required.");
       return;
@@ -171,11 +173,11 @@ public class AsyncLocalStorageUtil {
 
     try {
       ReactDatabaseSupplier supplier = ReactDatabaseSupplier.getInstance(ctx);
-      supplier.get().execSQL("PRAGMA wal_checkpoint");
+      supplier.get().rawQuery("PRAGMA wal_checkpoint", null).close();
       supplier.closeDatabase();
       Log.i("AsyncStorage_Next", "Forcing SQLite checkpoint successful.");
     } catch (Exception e) {
-      Log.w("AsyncStorage_Next", "Could not force checkpoint on RKStorage: " + e.getMessage());
+      Log.w("AsyncStorage_Next", "Could not force checkpoint on RKStorage, the Next storage might not migrate the data properly: " + e.getMessage());
     }
   }
 }
--- a/android/src/main/java/com/reactnativecommunity/asyncstorage/AsyncLocalStorageUtil.java
+++ b/android/src/main/java/com/reactnativecommunity/asyncstorage/AsyncLocalStorageUtil.java
@@ -8,11 +8,9 @@
 package com.reactnativecommunity.asyncstorage;
 
 import javax.annotation.Nullable;
-
 import java.io.File;
 import java.util.Arrays;
 import java.util.Iterator;
-
 import android.content.ContentValues;
 import android.content.Context;
 import android.database.Cursor;
@@ -20,12 +18,9 @@ import android.database.sqlite.SQLiteDatabase;
 import android.os.Build;
 import android.text.TextUtils;
 import android.util.Log;
-
 import com.facebook.react.bridge.ReadableArray;
-
 import org.json.JSONException;
 import org.json.JSONObject;
-
 import static com.reactnativecommunity.asyncstorage.ReactDatabaseSupplier.KEY_COLUMN;
 import static com.reactnativecommunity.asyncstorage.ReactDatabaseSupplier.TABLE_CATALYST;
 import static com.reactnativecommunity.asyncstorage.ReactDatabaseSupplier.VALUE_COLUMN;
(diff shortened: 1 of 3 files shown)


android/src/main/java/com/reactnativecommunity/asyncstorage/AsyncLocalStorageUtil.java
```diff
@@ -142,4 +146,36 @@ private static void deepMergeInto(JSONObject oldJSON, JSONObject newJSON)
...
+   * Because of that, migration to the next storage might not be successful, because the content of
+   * RKStorage might be still in WAL file instead. Committing all data from WAL to db file is called
+   * a "checkpoint" and is done automatically (by default) when the WAL file reaches a threshold
+   * size of 1000 pages.
+   * More here: https://sqlite.org/wal.html
+   * <p>
+   * This helper will force checkpoint on RKStorage, if Next storage file does not exists yet.
+   */
+  public static void verifyAndForceSqliteCheckpoint(Context ctx) {
+    File nextStorageFile = ctx.getDatabasePath("AsyncStorage");
```
Krizzu, line 162: I'm aware that this leaks next storage DB name, but I think it's harmless here (and db file name does not change much 😄 )

android/src/main/java/com/reactnativecommunity/asyncstorage/AsyncLocalStorageUtil.java
```diff
@@ -142,4 +146,36 @@ private static void deepMergeInto(JSONObject oldJSON, JSONObject newJSON)
...
+    File currentStorageFile = ctx.getDatabasePath(ReactDatabaseSupplier.DATABASE_NAME);
+    boolean isCheckpointRequired = Build.VERSION.SDK_INT >= Build.VERSION_CODES.P
```
tido64, line 164: Can we check `Build.VERSION.SDK_INT >= Build.VERSION_CODES.P` first at the top, then do the remaining checks? That way we don't do unnecessary work if the device is below Pie, even though it's probably not much in this case. But consistency is nice.

android/src/main/java/com/reactnativecommunity/asyncstorage/AsyncLocalStorageUtil.java
```diff
@@ -142,4 +146,36 @@ private static void deepMergeInto(JSONObject oldJSON, JSONObject newJSON)
...
+      return;
+    }
+
+    try {
+      ReactDatabaseSupplier supplier = ReactDatabaseSupplier.getInstance(ctx);
+      supplier.get().execSQL("PRAGMA wal_checkpoint");
+      supplier.closeDatabase();
+      Log.i("AsyncStorage_Next", "Forcing SQLite checkpoint successful.");
+    } catch (Exception e) {
+      Log.w("AsyncStorage_Next", "Could not force checkpoint on RKStorage: " + e.getMessage());
```
tido64, line 178: Should we also add that we cannot guarantee that the migrated db may not contain the latest?

android/src/main/java/com/reactnativecommunity/asyncstorage/AsyncStoragePackage.java
```diff
@@ -1,19 +1,21 @@
 /**
  * Copyright (c) Facebook, Inc. and its affiliates.
- *
+ * <p>
```
tido64, line 3: Intentional?

android/src/main/java/com/reactnativecommunity/asyncstorage/AsyncStoragePackage.java
```diff
@@ -29,6 +31,7 @@
                 Class storageClass = Class.forName("com.reactnativecommunity.asyncstorage.next.StorageModule");
                 NativeModule inst = (NativeModule) storageClass.getDeclaredConstructor(new Class[]{ReactContext.class}).newInstance(reactContext);
                 moduleList.add(inst);
+                AsyncLocalStorageUtil.verifyAndForceSqliteCheckpoint(reactContext);
```
tido64, line 34: Shouldn't this line be added wherever we're doing the migration? Won't this force a checkpoint on every startup?

Krizzu: Yes, doing this here will basically check to do the checkpoint on every app startup. Ideally, we should do it only when the migration will be performed, but from what I checked, the migration process is fired after the new db file is created - meaning that forcing the checkpoint no longer make sense, since the old db file was already read.

I think it's fine to leave it here - we make sure we fire checkpoint only once, by checking if existence states of both db files.

Krizzu:
@tido64 thanks, ready for re-review 🙏

android/src/main/java/com/reactnativecommunity/asyncstorage/AsyncLocalStorageUtil.java
```diff
@@ -142,4 +146,38 @@ private static void deepMergeInto(JSONObject oldJSON, JSONObject newJSON)
...
+   * Because of that, migration to the next storage might not be successful, because the content of
+   * RKStorage might be still in WAL file instead. Committing all data from WAL to db file is called
+   * a "checkpoint" and is done automatically (by default) when the WAL file reaches a threshold
+   * size of 1000 pages.
+   * More here: https://sqlite.org/wal.html
+   *
+   * This helper will force checkpoint on RKStorage, if Next storage file does not exists yet.
+   */
+  public static void verifyAndForceSqliteCheckpoint(Context ctx) {
+    if(Build.VERSION.SDK_INT < Build.VERSION_CODES.P) {
```
tido64, line 162: Suuuuper nitpicky 😄 
```suggestion
    if (Build.VERSION.SDK_INT < Build.VERSION_CODES.P) {
```

android/src/main/java/com/reactnativecommunity/asyncstorage/AsyncStoragePackage.java
tido64: I still see the `<p>`?
```suggestion
 *
```

tido64 (approved):
LGTM 🚀

android/src/main/java/com/reactnativecommunity/asyncstorage/AsyncLocalStorageUtil.java
tido64: Maybe we can make this configurable some time in the future?

android/src/main/java/com/reactnativecommunity/asyncstorage/AsyncLocalStorageUtil.java
Krizzu: You got it 😄

android/src/main/java/com/reactnativecommunity/asyncstorage/AsyncLocalStorageUtil.java
Krizzu: Yeah, probably something to think about

android/src/main/java/com/reactnativecommunity/asyncstorage/AsyncStoragePackage.java
Krizzu: just my linter being smart 🗡️

PR #615, opened 2021-05-28
remove buildToolsVersion

dulmandakh:
## Summary

Setting buildToolsVersion is no longer recommended, it'll default to Android Gradle Plugin defaults. It makes DX worse by requiring multiple versions of build tools by various dependencies require various versions.

## Test Plan

No visible changes, everything will work as before.


Krizzu (approved):
Yup, thanks 🙏

PR #659, opened 2021-08-18
fix: declare support for react-native 0.65

### LINKED ISSUE #658: Compatibility with React Native 0.65
## You want to:  
Hi, yesterday a new version of RN was made publicly available, is the current version 1.15.6 compatible? Because there is a warning when installing 
`warning " > @react-native-async-storage/async-storage@1.15.6" has incorrect peer dependency "react-native@^0.60.6 || ^0.61.5 || ^0.62.2 || ^0.63.2 || ^0.64.0 || 1000.0.0".`

Can I just ignore it for now?

tido64:
## Summary

Declares support for react-native 0.65 and for react-native-macos/-windows canaries.

Also removes `resolutions` since metro dependencies should be in sync now.

Resolves #658.

## Test Plan

There's nothing to test.

### DIFF
--- a/package.json
+++ b/package.json
@@ -64,7 +64,7 @@
     "merge-options": "^3.0.4"
   },
   "peerDependencies": {
-    "react-native": "^0.60.6 || ^0.61.5 || ^0.62.2 || ^0.63.2 || ^0.64.0 || 1000.0.0"
+    "react-native": "^0.0.0-0 || ^0.60.6 || ^0.61.5 || ^0.62.2 || ^0.63.2 || ^0.64.0 || ^0.65.0 || 1000.0.0"
   },
   "devDependencies": {
     "@babel/core": "^7.8.4",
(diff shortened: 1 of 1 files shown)


package.json
```diff
@@ -64,7 +64,7 @@
     "merge-options": "^3.0.4"
   },
   "peerDependencies": {
-    "react-native": "^0.60.6 || ^0.61.5 || ^0.62.2 || ^0.63.2 || ^0.64.0 || 1000.0.0"
+    "react-native": "^0.0.0-0 || ^0.60.6 || ^0.61.5 || ^0.62.2 || ^0.63.2 || ^0.64.0 || ^0.65.0 || 1000.0.0"
```
arabold, line 67: I never fully understood why libraries need to be so specific about their support. From my understanding, `react-native-async-storage/async-storage` supports React Native 0.60.6 and above. So, why not simply use `"react-native": ">=0.60.6"` here instead?

Of course, we don't know whether it will support 0.66 and beyond yet, but I guess it's weighing the pros and cons of having to update the library every couple of months by adding a new version, versus being optimistic considering that there are only very few dependencies that actually _could_ break. I would rather accept the risk of things breaking in the future due to an unexpected change in React Native, than having to rely on the original author of a library continuously keeping it up to date. 

Or in other words (_in a dramatic voice_): It's already been three days, why hasn't this been merged yet? Whyyyy? Whyyyyyyy?

VitaliyBelishov: pls approve and marge it

Krizzu (approved):
Sorry for the delay 🙏

Issue #587, opened 2021-04-12
Setup async-storage for macOS


nikgraf:
## Current behavior

I followed the steps here https://react-native-async-storage.github.io/async-storage/docs/link/#macos

Everything until the 4th step is clear. Then I added the `libRNCAsyncStorage-macOS.a` in the `Build Phases`
of the target `desktop-macOS`.

When I run ` npx react-native run-macos` I get this error. What's odd is is that this is referencing the ios directory.

```
/node_modules/@react-native-async-storage/async-storage/ios/RNCAsyncStorage.h:10:9: fatal error: 'React/RCTBridgeModule.h' file not found
#import <React/RCTBridgeModule.h>
        ^~~~~~~~~~~~~~~~~~~~~~~~~
1 error generated.
```

## Expected behavior

Get 

## Repro steps

Can provide an example repo in case the above description doesn't help.

## Environment

- **Platforms tested**:
  - [ ] Android
  - [ ] iOS
  - [x] macOS
  - [ ] Windows
- **AsyncStorage version**: 1.15.2
- **Expo version**: 
- **Environment**:

```
System:
    OS: macOS 11.1
    CPU: (8) x64 Apple M1
    Memory: 294.54 MB / 16.00 GB
    Shell: 5.8 - /bin/zsh
  Binaries:
    Node: 14.15.1 - ~/.nvm/versions/node/v14.15.1/bin/node
    Yarn: 1.22.10 - /usr/local/bin/yarn
    npm: 6.14.8 - ~/.nvm/versions/node/v14.15.1/bin/npm
    Watchman: Not Found
  Managers:
    CocoaPods: 1.10.1 - /usr/local/bin/pod
  SDKs:
    iOS SDK:
      Platforms: iOS 14.4, DriverKit 20.2, macOS 11.1, tvOS 14.3, watchOS 7.2
    Android SDK: Not Found
  IDEs:
    Android Studio: Not Found
    Xcode: 12.4/12D4e - /usr/bin/xcodebuild
  Languages:
    Java: Not Found
    Python: 2.7.16 - /usr/bin/python
  npmPackages:
    @react-native-community/cli: Not Found
    react: 16.13.1 => 16.13.1 
    react-native: 0.63.4 => 0.63.4 
    react-native-macos: ^0.63.0-0 => 0.63.23 
  npmGlobalPackages:
    *react-native*: Not Found
```

- **Logs/Errors that may be relevant**:

tido64:
[...]

PR #678, opened 2021-09-22
docs(website): update auto-linking info for macOS

mmazzarolo:
Since React Native for macOS 0.63, autolink works out-of-the-box as long as `platform :macos, '10.14'` is set in Podfile)

See #587

### DIFF
--- a/website/docs/Installation.md
+++ b/website/docs/Installation.md
@@ -51,4 +51,10 @@ react-native link @react-native-async-storage/async-storage
 For earlier versions check [manual linking](Linking.md)
 
 #### macOS
-For `macOS` the [manual linking](Linking.md) is currently the only linking option.
+- **React Native macOS >= 0.63**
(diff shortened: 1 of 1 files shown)


mmazzarolo:
I assume CI failing here is not related to the PR changes, right?

Krizzu:
No, it's not. CircleCI has some issues currently, so maybe that's why Android unit test fails. Regarding Windows, maybe @tido64 would have an idea.

Anyway, I'm happy to merge it - @tido64 might having a look as well? 🙏🏼

tido64:
> No, it's not. CircleCI has some issues currently, so maybe that's why Android unit test fails. Regarding Windows, maybe @tido64 would have an idea.

No idea what's going on. Looks like it could be some caching issue? Not seeing this on master. In any case, let's just force merge this while we figure it out.

PR #679, opened 2021-09-26
Upgrade Kotlin to 1.5 and Room to 2.3.0

Minishlink:
## Summary

My app uses Kotlin 1.5 and I cannot downgrade to 1.4, so I upgraded locally to 1.5. It appeared I also needed to upgrade Room to the last stable version.
I'm not sure if this is a breaking change, or maybe it would be better to let users choose their room version.

### DIFF
--- a/android/build.gradle
+++ b/android/build.gradle
@@ -35,7 +35,7 @@ buildscript {
             ? rootProject.ext['kotlinVersion']
             : rootProject.hasProperty('AsyncStorage_kotlinVersion')
             ? rootProject.properties['AsyncStorage_kotlinVersion']
-            : '1.4.21'
+            : '1.5.31'
 
     repositories {
         google()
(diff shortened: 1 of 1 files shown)


Krizzu (changes requested):
Hey @Minishlink 

I haven't tested it against latest Kotlin/Room, so I cannot approve the PR. However, there's a way for you to provide a Kotlin version (see [storage documentation, section "How to specifying Kotlin version"](https://react-native-async-storage.github.io/async-storage/docs/advanced/next#enabling). 

Regarding Room - I think it's a good idea to provide this flexibility too. I'd be happy to accept a PR with this proposal.

In meanwhile, if you're looking for a way to keep versions you specified in your project, consider using [patch-package library](https://www.npmjs.com/package/patch-package) that could help with that

Issue #684, opened 2021-10-07
Issue on M1 macOS with current room version


Daavidaviid:
## Current behavior

It crashes during native build time, but I was able to solve it using forcing the room version to 2.4.0-alpha03
https://stackoverflow.com/questions/68922875/room-database-is-not-working-in-mac-book-pro-m1

## Environment

- **Platforms tested**:
  - [X] Android
  - [ ] iOS
  - [ ] macOS
  - [ ] Windows
- **AsyncStorage version**: 
- **Expo version**: 
- **Environment**:

version 1.15.8
AsyncStorage_useNextStorage=true

- **Logs/Errors that may be relevant**: 

```bash
* What went wrong:
Execution failed for task ':react-native-async-storage_async-storage:kaptDebugKotlin'.
> A failure occurred while executing org.jetbrains.kotlin.gradle.internal.KaptWithoutKotlincTask$KaptExecutionWorkAction
   > java.lang.reflect.InvocationTargetException (no error message)
```

tido64:
Resolved by #698

samithaFHSS:
This is not yet resolved. Still getting the same error on `@react-native-async-storage/async-storage": "^1.15.13"` because room 2.3.0 does not support Apple M1 silicone.
I tried to use `AsyncStorage_next_roomVersion` to define the beta room version but it generates another error. Would appreciate if you can provide a solution for this.

Added below line to gradle.properties
```
AsyncStorage_next_roomVersion="2.4.0-beta02"
```

New error
```
Could not resolve androidx.room:room-runtime:"2.4.0-beta02".
         > Could not get resource 'https://jcenter.bintray.com/androidx/room/room-runtime/%222.4.0-beta02%22/room-runtime-%222.4.0-beta02%22.pom'.
            > Could not GET 'https://jcenter.bintray.com/androidx/room/room-runtime/%222.4.0-beta02%22/room-runtime-%222.4.0-beta02%22.pom'. Received status code 400 from server: Bad Request

```

Issue #634, opened 2021-07-15
[Android] multiGet() hangs, app not responsive


ecexplorer:
## Current behavior

In our app, I call `AsyncStorage.multiGet()` on a key array with just two keys. Admittedly one of the keys is tied to a JSON string that when formatted, is 17000+ lines long. This is still under the overall 6MB limit.  
When the call is made, it just hangs. No error is thrown, and if I use a callback, it is never called. The same data in iOS has no problem being fetched.  
I can also tell that the data is there. That 17,000+ number comes from pulling the data directly on Android Studio using the database inspector. The data is being stored correctly.  
If I request the same data using `getItem()`, then the request works without issue and I get the expected value back.  
(See comments for a workaround)

## Expected behavior

Calls to `multiGet()` don't hang, return the data expected.

## Repro steps

- Set two items in AsyncStorage, ensure one of those is LARGE.
- Attempt to get both at once with a `multiGet()` request.
- If the request hangs, this has been repeated.  

Similar to source, if I make a Promise.all(keys.map(async (key) => [key, await AsyncStorage.getItem(key)]), I see the same hang. Android doesn't seem to like the multiple accesses to AsyncStorage/DB at once.

## Environment

- **Platforms tested**:
  - [ X] Android
  - [ ] iOS
  - [ ] macOS
  - [ ] Windows
- **AsyncStorage version**: 1.15.5
- **Environment**:
 System:
    OS: macOS 11.4
    CPU: (16) x64 Intel(R) Core(TM) i9-9880H CPU @ 2.30GHz
    Memory: 131.27 MB / 32.00 GB
    Shell: 5.8 - /bin/zsh
  Binaries:
    Node: 12.16.2 - ~/.nvm/versions/node/v12.16.2/bin/node
    Yarn: Not Found
    npm: 6.14.4 - ~/.nvm/versions/node/v12.16.2/bin/npm
    Watchman: 2021.06.07.00 - /usr/local/bin/watchman
  Managers:
    CocoaPods: 1.10.1 - /usr/local/bin/pod
  SDKs:
    iOS SDK:
      Platforms: iOS 14.5, DriverKit 20.4, macOS 11.3, tvOS 14.5, watchOS 7.4
    Android SDK: Not Found
  IDEs:
    Android Studio: 4.2 AI-202.7660.26.42.7486908
    Xcode: 12.5.1/12E507 - /usr/bin/xcodebuild
  Languages:
    Java: Not Found
[...]

PR #698, opened 2021-10-25
fix: Multiple improvements + docs update

Krizzu:
## Summary

Sorry for pushing it into one PR. Got a couple of fixes/improvements coming:

#### Fixed the issue where `multiGet` did not propagate errors correctly, meaning they were just ignored. #634

This also changes the way it handles the error from original design - previously, if we'd call many `multiGet` for different keys, they'll be batched into one native call. If any of them failed, the rest of them fail too, but, only the failed one will get the error message, while the rest will be resolved as `undefined` values, creating false negatives. This is now changed, so that once one multiGet fails, every other batched call will throw the same error.

#### Bumped the dependencies for Next storage (Kotlin, Room, Coroutines). #684
I tried to get everything to the latest version. Turns out, Room, with version 2.4.0 (where Apple M1 support is fixed) needs also increased `compileSdk` version bumped to 31 (latest one). It'd be bad for a library to dictate this version, so instead hardcoding it to 2.4.0, I made it 2.3.0. However, I added a `feature flag`, so that developer can set this version manually. (cc @Daavidaviid) 

#### Updated documentation

Since we had a few issues raised about size limits, I added a section in our documentation about it.

### DIFF
--- a/package.json
+++ b/package.json
@@ -88,7 +88,7 @@
     "react-native": "0.63.4",
     "react-native-builder-bob": "^0.18.0",
     "react-native-macos": "^0.63.4",
-    "react-native-test-app": "^0.9.0",
+    "react-native-test-app": "^0.9.5",
     "react-native-web": "~0.12.0",
     "react-native-windows": "^0.63.41",
     "react-test-renderer": "16.13.1",
--- a/src/AsyncStorage.native.js
+++ b/src/AsyncStorage.native.js
@@ -297,13 +297,12 @@ const AsyncStorage = {
 
       for (let i = 0; i < reqLength; i++) {
         const request = getRequests[i];
-        const requestKeys = request.keys;
         if (error) {
           request.callback && request.callback(error);
           request.reject && request.reject(error);
           continue;
         }
-        const requestResult = requestKeys.map((key) => [key, map[key]]);
+        const requestResult = request.keys.map((key) => [key, map[key]]);
         request.callback && request.callback(null, requestResult);
         request.resolve && request.resolve(requestResult);
       }
--- a/src/AsyncStorage.native.js
+++ b/src/AsyncStorage.native.js
@@ -282,9 +282,27 @@ const AsyncStorage = {
           return value;
         });
       const reqLength = getRequests.length;
+
+      /**
+       * As mentioned few lines above, this method could be called with the array of potential error,
+       * in case of anything goes wrong. The problem is, if any of the batched calls fails
+       * the rest of them would fail too, but the error would be consumed by just one. The rest
+       * would simply return `undefined` as their result, rendering false negatives.
+       *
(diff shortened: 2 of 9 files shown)


src/AsyncStorage.native.js
```diff
@@ -282,9 +282,27 @@ const AsyncStorage = {
...
       const reqLength = getRequests.length;
+
+      /**
+       * As mentioned few lines above, this method could be called with the array of potential error,
+       * in case of anything goes wrong. The problem is, if any of the batched calls fails
+       * the rest of them would fail too, but the error would be consumed by just one. The rest
+       * would simply return `undefined` as their result, rendering false negatives.
+       *
+       * In order to avoid this situation, in case of any call failing,
+       * the rest of them will be rejected as well (with the same error).
```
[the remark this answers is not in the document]
Krizzu: Yes, I do agree. This change here will do exactly that, for each platform

Krizzu:
It's ready for re-review. I haven't touch the code regarding error handling, since it handles all platforms equally. If you'd rather not throw the same error for every batched `multiGet` call, let me know.

src/AsyncStorage.native.js
```diff
@@ -282,9 +282,27 @@ const AsyncStorage = {
...
+
       for (let i = 0; i < reqLength; i++) {
         const request = getRequests[i];
         const requestKeys = request.keys;
+        if (error) {
+          request.callback && request.callback(error);
+          request.reject && request.reject(error);
+          continue;
+        }
         const requestResult = requestKeys.map((key) => [key, map[key]]);
```
tido64, line 306: We can move `requestKeys` below the if-check since it doesn't get used here:
```suggestion
        if (error) {
          request.callback && request.callback(error);
          request.reject && request.reject(error);
          continue;
        }

        const requestResult = request.keys.map((key) => [key, map[key]]);
```

tido64 (approved):
Just a single nit. Looks good otherwise.

package.json
```diff
@@ -88,7 +88,7 @@
     "react-native": "0.63.4",
     "react-native-builder-bob": "^0.18.0",
     "react-native-macos": "^0.63.4",
-    "react-native-test-app": "^0.9.0",
+    "react-native-test-app": "^0.9.5",
```
Krizzu, line 91: @tido64 I've noticed that `react-native-test-app` is overriding `kotlinVersion` from root project, which leads to build fails if I override it myself. Looked up the release history of test-app, I updated to 0.9.5, where Kotlin was bumped to desired version.

Are there any extra steps needed to this version bump?

tido64: No, it should just work. Are you seeing issues?

Krizzu: Nope, just making sure 😄

schmod:
Where are you seeing that Room 2.4.0 requires compileSdk 31?

PR #706, opened 2021-11-04
fix: Gradle build failing for version 1.15.10

Krizzu:
## Summary

Could reproduce this in a clean project, but not on our example app. Because method used inside `buildscript` is not reachable, I instead read properties directly.


Krizzu:
@tido64 I'm putting this on auto-merge, since it blocks people from using AS

tido64 (approved):

PR #707, opened 2021-11-08
Make web version work without unknown env vars set

arthurfranca:
Web doesn't have process.env.NODE_ENV nor __DEV __ vars set.
Checking for these vars will raise an `Uncaught ReferenceError`.

In my case, i use webpack to replace process.env.NODE_ENV with a value at build time, but the `process && ...` check would still raise an error (and replacing process object [isn't good, see warning on webpack docs](https://webpack.js.org/plugins/define-plugin/))

### DIFF
--- a/packages/core/src/defaults.ts
+++ b/packages/core/src/defaults.ts
@@ -10,7 +10,7 @@ import {FactoryOptions, LoggerAction} from '../types';
 
 const isDevelopment =
   // @ts-ignore
-  (process && process.env && process.env.NODE_ENV === 'development') || __DEV__;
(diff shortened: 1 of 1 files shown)


Krizzu:
what kind of error you see with `process && process.env ...` check? If `process` or `__DEV__` does not exist, it'd just fallback to `undefined`, right?

arthurfranca:
> what kind of error you see with `process && process.env ...` check? If `process` or `__DEV__` does not exist, it'd just fallback to `undefined`, right?

When one tries to access an undefined variable, it won't fallback to `undefined` value.  It will raise an error, which will prevent any script after the lib first import from running.

Krizzu:
Ah, of course! I thought about completely something else 😄

Krizzu (approved):
Although this PR is legit and we could merge this, v2 won't be continued in the current shape.

PR #719, opened 2021-11-28
fix(windows): Fix crash when running windows module

vmoroz:
## Summary

Lately we saw a number of Watson crash reports in Windows that show crashes due to the result callbacks called twice in case if there are errors related to the database operations. The analysis has shown that the code has no protection from the callbacks being called twice and that the `multiMerge` method is not implemented in a safe way as other methods intended to be.

## Details
This PR is relatively big. It is almost a full rewrite of Windows module code. These are the changes:

- To ensure that the result callbacks are called exactly once (not more or less than once) we are adding `Promise` class that invokes the result callback when the `Promise` is resolved or rejected. To ensure that the callback is called not more than once, it uses `m_isCompleted` Boolean atomic variable. To ensure that the callback is called at least once, the `Reject` method is called from destructor.
- To implement strongly-typed module API, the new `KeyValue` and `Error` classes are added. These new types have custom `ReadValue` and `WriteValue` function overloads to convert to/from JSON. It allowed to stop using `JSValue` in favor of the strongly typed classes such as `std::vector`, `std::string`, `KeyValue`, and `Error`.
- The Database initialization is moved from constructor to the `InitializeStorage` method that allows to avoid using exceptions and to report errors the same way as other async task methods. This method is called on demand when a task started.
- All error reporting was changed to uniformly add errors to the error list associated with the `Promise` and indicate error result by the `std::nullopt` value. Most of all functions return `std::optional`.
- Convenience macros `CHECK` and `CHECK_SQL_OK` are added to simplify code by reducing `if-return` statements for each function call.
- `UniquePtrSetter` is added to allow natural use of `std::unique_ptr` for SQL object handlers.
- Since the updated `DBTask` is not copyable, the task list changed to hold `std::unique_ptr<DBTask` instances.
- Implemented `MultiMerge` method as other methods inside of `DBTask`. We needed to fix algorithm that matches keys between old and new values to use `std::unordered_set` instead of relying on their index which may be different and cause incorrect merges.
- Fixed `MergeJsonObjects` code to always check value type and that key exists.
- Cleaned up the `Sqlite3Transaction` class code.
- Fixed Microsoft copyright notice to follow the official company guidelines.

## Test Plan

The code is manually tested using the example app. The example app is generated using the following command:
```
yarn install-windows-test-app -p example/windows
```

### DIFF
--- a/windows/code/DBStorage.cpp
+++ b/windows/code/DBStorage.cpp
@@ -22,7 +22,7 @@ namespace winrt
     }
 
 // Convenience macro to call CheckSQLiteResult.
-#define CHECK_SQL_OK(expr) CHECK(CheckSQLiteResult(db, m_errorHandler, (expr)))
+#define CHECK_SQL_OK(expr) CHECK(CheckSQLiteResult(db, m_errorManager, (expr)))
 
 namespace
 {
@@ -60,9 +60,9 @@ namespace
                                                 char **columnNames);
 
     // Execute the provided SQLite statement (and optional execCallback & user data
-    // in callbackData). On error, report it to the errorHandler and return std::nullopt.
+    // in callbackData). On error, report it to the errorManager and return std::nullopt.
     std::optional<bool> Exec(sqlite3 *db,
-                             DBStorage::ErrorHandler &errorHandler,
+                             DBStorage::ErrorManager &errorManager,
                              const char *statement,
                              ExecCallback execCallback = nullptr,
                              void *callbackData = nullptr) noexcept
@@ -70,10 +70,10 @@ namespace
         auto errMsg = std::unique_ptr<char, decltype(&sqlite3_free)>{nullptr, &sqlite3_free};
         int rc = sqlite3_exec(db, statement, execCallback, callbackData, &errMsg);
         if (errMsg) {
-            return errorHandler.AddError(errMsg.get());
+            return errorManager.AddError(errMsg.get());
         }
         if (rc != SQLITE_OK) {
-            return errorHandler.AddError(sqlite3_errmsg(db));
+            return errorManager.AddError(sqlite3_errmsg(db));
         }
         return true;
     }
@@ -81,11 +81,11 @@ namespace
     // Convenience wrapper for using Exec with lambda expressions.
     template <typename Fn>
     std::optional<bool>
-    Exec(sqlite3 *db, DBStorage::ErrorHandler &errorHandler, const char *statement, Fn &fn) noexcept
+    Exec(sqlite3 *db, DBStorage::ErrorManager &errorManager, const char *statement, Fn &fn) noexcept
     {
         return Exec(
             db,
-            errorHandler,
+            errorManager,
             statement,
             [](void *callbackData, int columnCount, char **columnTexts, char **columnNames) {
                 return (*static_cast<Fn *>(callbackData))(columnCount, columnTexts, columnNames);
@@ -95,9 +95,9 @@ namespace
 
     // Check that the args collection size is less than SQLITE_LIMIT_VARIABLE_NUMBER, and that
     // every member of args is not an empty string.
-    // On error, report it to the errorHandler and return std::nullopt.
+    // On error, report it to the errorManager and return std::nullopt.
     std::optional<bool> CheckArgs(sqlite3 *db,
-                                  DBStorage::ErrorHandler &errorHandler,
+                                  DBStorage::ErrorManager &errorManager,
                                   const std::vector<std::string> &args) noexcept
     {
         int varLimit = sqlite3_limit(db, SQLITE_LIMIT_VARIABLE_NUMBER, -1);
@@ -106,11 +106,11 @@ namespace
             static_cast<int>(argCount) > varLimit) {
             char errorMsg[60];
             sprintf_s(errorMsg, "Too many keys. Maximum supported keys :%d.", varLimit);
-            return errorHandler.AddError(errorMsg);
+            return errorManager.AddError(errorMsg);
         }
         for (int i = 0; i < static_cast<int>(argCount); i++) {
             if (args[i].empty()) {
-                return errorHandler.AddError("The key must be a non-empty string.");
+                return errorManager.AddError("The key must be a non-empty string.");
             }
         }
         return true;
@@ -118,13 +118,13 @@ namespace
 
     // RAII object to manage SQLite transaction. On destruction, if
     // Commit() has not been called, rolls back the transactions.
-    // The provided SQLite connection handle & errorHandler must outlive
+    // The provided SQLite connection handle & errorManager must outlive
     // the Sqlite3Transaction object.
     struct Sqlite3Transaction {
-        Sqlite3Transaction(sqlite3 *db, DBStorage::ErrorHandler &errorHandler) noexcept
-            : m_db(db), m_errorHandler(errorHandler)
+        Sqlite3Transaction(sqlite3 *db, DBStorage::ErrorManager &errorManager) noexcept
+            : m_db(db), m_errorManager(errorManager)
         {
-            if (!Exec(m_db, m_errorHandler, "BEGIN TRANSACTION")) {
+            if (!Exec(m_db, m_errorManager, "BEGIN TRANSACTION")) {
                 m_db = nullptr;
             }
         }
@@ -145,7 +145,7 @@ namespace
         std::optional<bool> Commit() noexcept
         {
             if (m_db) {
-                return Exec(std::exchange(m_db, nullptr), m_errorHandler, "COMMIT");
+                return Exec(std::exchange(m_db, nullptr), m_errorManager, "COMMIT");
             }
             return std::nullopt;
         }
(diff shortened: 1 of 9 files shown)


asklar (approved):
LGTM, left a few comments. Thanks for taking the time to rewrite it, looks cleaner now : )

windows/code/DBStorage.h
```diff
@@ -1,71 +1,160 @@
...
+        {
+        }
+
+        ~Promise()
+        {
+            Reject();
+        }
+
+        Promise(const Promise &other) = delete;
+        Promise &operator=(const Promise &other) = delete;
```
tido64, line 49: Do we also need to delete the `operator=`s?

windows/code/DBStorage.cpp
```diff
@@ -187,160 +177,158 @@ namespace
...
-            m_mutex.lock_shared();
-        }
-
-        ~slim_shared_lock_guard() noexcept
-        {
-            m_mutex.unlock_shared();
+    // Merge source into destination.
+    // It only merges objects - all other types are just clobbered (including arrays).
+    void MergeJsonObjects(winrt::Windows::Data::Json::JsonObject const &destination,
+                          winrt::Windows::Data::Json::JsonObject const &source) noexcept
```
tido64, line 212: Nit: Use the aliases you created earlier?
```suggestion
    void MergeJsonObjects(winrt::JsonObject const &destination,
                          winrt::JsonObject const &source) noexcept
```

windows/code/DBStorage.cpp
```diff
@@ -351,174 +339,236 @@ winrt::Windows::Foundation::IAsyncAction DBStorage::RunTasks()
...
+{
+    if (!db) {
+        // We initialize DB handler on demand to report errors in the task context.
+        if (auto res = storage.InitializeStorage(m_errorHandler)) {
+            db = *res;
+        }
+    }
+    if (db) {
+        m_onRun(*this, db);
     }
```
tido64, line 385: Should we log an error message if there is no db here or will that be too spammy?

windows/code/DBStorage.cpp
```diff
@@ -351,174 +339,236 @@ winrt::Windows::Foundation::IAsyncAction DBStorage::RunTasks()
...
+    return true;
+}
+
+// Read KeyValue from a JSON array.
+void ReadValue(const winrt::IJSValueReader &reader,
+               /*out*/ DBStorage::KeyValue &value) noexcept
+{
+    if (reader.ValueType() == winrt::JSValueType::Array) {
+        int index = 0;
+        while (reader.GetNextArrayItem()) {
```
tido64, line 544: For my own education, how does this work? It looks like we have to call `GetNextArrayItem()` at least twice, once for the key and again for the value. Then for some reason, we have to keep calling `GetNextArrayItem()` until it returns `false`? Why is that?

tido64 (approved):
@Krizzu: Do you want to take a look?

Krizzu:
> @Krizzu: Do you want to take a look?

I'll leave it with you 👍  😄

windows/code/DBStorage.h
```diff
@@ -1,71 +1,160 @@
...
+
+        ErrorHandler &GetErrorHandler() noexcept
```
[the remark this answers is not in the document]
vmoroz: No, it is intentional: we use this method to share the ErrorHandler instance with the Async task.

windows/code/DBStorage.h
```diff
@@ -1,71 +1,160 @@
...
+        ErrorHandler m_errorHandler;
+        std::atomic_flag m_isCompleted{false};
+        TOnResolve m_onResolve;
+        TOnReject m_onReject;
+    };
+
+    // An asynchronous task that run in a background thread.
+    struct DBTask {
+        DBTask(ErrorHandler &errorHandler,
+               std::function<void(DBTask &task, sqlite3 *db)> &&onRun) noexcept;
```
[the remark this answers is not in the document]
vmoroz: The repo has .clang-format configuration that defines the alignment rules. If we want to change, then it must be a different PR.

windows/code/RNCAsyncStorage.h
```diff
@@ -1,175 +1,110 @@
...
-                        if (callbackParams.size() > 1) {
-                            callback(error, callbackParams[1].AsArray());
-                        } else {
-                            callback(error, {});
-                        }
-                    }
+            auto promise = DBStorage::CreatePromise(
+                [callback](const std::vector<std::string> &keys) { callback(std::nullopt, keys); },
+                [callback](const std::vector<DBStorage::Error> &errors) {
+                    callback(errors.at(0), {});
```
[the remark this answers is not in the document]
vmoroz: Good point! I am going to concatenate them instead.

windows/code/DBStorage.cpp
```diff
@@ -351,174 +339,236 @@ winrt::Windows::Foundation::IAsyncAction DBStorage::RunTasks()
...
+                ReadValue(reader, value.Key);
+            } else if (index == 1) {
+                ReadValue(reader, value.Value);
+            } else {
+                // Read the array till the end to keep reader in a good state.
+                winrt::SkipValue<winrt::JSValue>(reader);
+            }
+            ++index;
+        }
     }
```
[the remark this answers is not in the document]
vmoroz: We cannot throw an error because these methods are noexcept and it will cause the app crash. I think we need to add a method to IJSValueReader to be able to report such errors. I have created [RNW #9216](https://github.com/microsoft/react-native-windows/issues/9216) for myself to track it.

windows/code/DBStorage.h
vmoroz: Yes, to ensure that the type is not copyable we must delete the copy constructor and the copy assignment. We do not want allow this class to be copyable because it has an atomic variable that ensures that we call Resolve or Reject only once. If could copy the class state, we would lose such guarantee.

windows/code/DBStorage.cpp
```diff
@@ -351,174 +339,236 @@ winrt::Windows::Foundation::IAsyncAction DBStorage::RunTasks()
...
+    if (reader.ValueType() == winrt::JSValueType::Array) {
+        int index = 0;
+        while (reader.GetNextArrayItem()) {
+            if (index == 0) {
+                ReadValue(reader, value.Key);
+            } else if (index == 1) {
+                ReadValue(reader, value.Value);
+            } else {
+                // Read the array till the end to keep reader in a good state.
+                winrt::SkipValue<winrt::JSValue>(reader);
```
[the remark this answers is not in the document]
vmoroz: We cannot do it today. I have created [RNW #9216](https://github.com/microsoft/react-native-windows/issues/9216) to track it.

windows/code/DBStorage.cpp
```diff
@@ -1,180 +1,170 @@
...
-            : m_db(db), m_callback(&callback)
+    // Commit() has not been called, rolls back the transactions.
+    // The provided SQLite connection handle & errorHandler must outlive
+    // the Sqlite3Transaction object.
+    struct Sqlite3Transaction {
+        Sqlite3Transaction(sqlite3 *db, DBStorage::ErrorHandler &errorHandler) noexcept
+            : m_db(db), m_errorHandler(errorHandler)
         {
-            if (!Exec(m_db, *m_callback, u8"BEGIN TRANSACTION")) {
+            if (!Exec(m_db, m_errorHandler, "BEGIN TRANSACTION")) {
```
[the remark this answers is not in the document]
vmoroz: Let me check this project with C++20.

windows/code/DBStorage.h
tido64: For some reason, I thought we were _missing_ `operator=` here 😛  What about the move constructor/operator?

vmoroz: The move constructor/operator= are removed automatically when we remove the copy constructor/operator=. If we want them enable, then they must be added back explicitly. They can be either `=default` or custom.

windows/code/DBStorage.cpp
vmoroz: I am not sure if we log other error message. Let me check.

windows/code/DBStorage.cpp
vmoroz: You are right, we need to call it only twice for Key and Value. We do it in the cycle in case if data is wrong and someone gave us an array with an unknown number of items. This way the first item will be Key if it is present, the second - Value if it is present, and then we just ignore the rest. Ideally, we would avoid this dance, but I have no way to stop and report an error. This is definitely a flaw in the ReadValue design. I will think how to fix it in future versions of RNW and then we can change it here.

windows/code/DBStorage.cpp
asklar, line 551: by the same principle, should we skip in the outer clause, until we find an array (e.g. if the value we got isn't an array)?

vmoroz: Good point! I will fix it.

windows/code/DBStorage.cpp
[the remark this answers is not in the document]
vmoroz: No, it is not required for C++20. I found another issue though.

windows/code/DBStorage.cpp
vmoroz: I looked at the code. If there is no db here, then it means that we failed to initialize it earlier in the `InitializeStorage` method above. This method adds any related errors to the ErrorManager (replaced ErrorHandler in latest code.)

windows/code/DBStorage.h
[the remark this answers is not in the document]
vmoroz: I have added a new method GetCombinedError to concatenate all errors if we have more than one.

windows/code/DBStorage.cpp
vmoroz: It is fixed.

PR #720, opened 2021-11-29
chore: bump react-native to 0.64

tido64:
## Summary

Bumps react-native to 0.64. This only affects maintainers.

## Test Plan

CI should be green.


android/src/test/java/com/reactnativecommunity/asyncstorage/next/ArgumentHelpersTest.kt
```diff
@@ -1,79 +0,0 @@
-package com.reactnativecommunity.asyncstorage.next
```
Krizzu, line 1: why have you removed those tests?

android/src/test/java/com/reactnativecommunity/asyncstorage/next/TestUtils.kt
```diff
@@ -1,84 +0,0 @@
-package com.reactnativecommunity.asyncstorage.next
```
Krizzu, line 1: This is used by `ArgumentHelpersTest.kt`, why removing this?

android/src/test/java/com/reactnativecommunity/asyncstorage/next/ArgumentHelpersTest.kt
tido64: These are only testing `createNativeCallArguments`, and `createNativeCallArguments` is a test-only utility. They're not testing real code as far as I can see. Did I miss anything?

Krizzu: Might seem like it, but these test are actually checking [.toEntryList](https://github.com/react-native-async-storage/async-storage/blob/master/android/src/main/java/com/reactnativecommunity/asyncstorage/next/ArgumentHelpers.kt#L8) and  [.toKeyList](https://github.com/react-native-async-storage/async-storage/blob/master/android/src/main/java/com/reactnativecommunity/asyncstorage/next/ArgumentHelpers.kt#L33) extension functions

tido64: Aha! Let me restore them then.

android/src/test/java/com/reactnativecommunity/asyncstorage/next/TestUtils.kt
```diff
@@ -1,84 +1,19 @@
...
+import com.facebook.react.bridge.JavaOnlyArray
 import com.facebook.react.bridge.ReadableArray
-import com.facebook.react.bridge.ReadableMap
-import com.facebook.react.bridge.ReadableType
-import com.facebook.react.bridge.WritableArray
 
-fun <T : Any> createNativeCallArguments(vararg list: T): ReadableArray {
-    val args = ArrayList(list.toList())
-    return TestingArguments(args)
+fun List<Any?>.toReadableArray(): ReadableArray {
```
Krizzu, line 6: Nice one 👍 
You could actually move this extension to `ArgumentHelpersTest.kt` file, to remove this one

Krizzu (approved):
This looks awesome, well done 👍

android/src/test/java/com/reactnativecommunity/asyncstorage/next/TestUtils.kt
tido64: Yeah, I think that makes sense.

PR #728, opened 2021-12-23
Update build.gradle

Daavidaviid:
## Summary

Updating to latest room version `2.4.0` to improve compatibility with M1 macs.
Need `compileSdkVersion >= 31` to work

### DIFF
--- a/android/build.gradle
+++ b/android/build.gradle
@@ -115,7 +115,7 @@ repositories {
 dependencies {
 
     if (useNextStorage) {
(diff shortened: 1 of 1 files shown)


Krizzu:
Because Room version bump requires bump in compileSdk, I did not make it 2.4 by default. Instead, I added a way to specify the version in your project, [check out the configuration part here](https://react-native-async-storage.github.io/async-storage/docs/advanced/next#configuration)

PR #753, opened 2022-02-16
chore: bump semantic-release to 19.0

tido64:
## Summary

This will also address https://github.com/advisories/GHSA-rrrm-qjm4-v8hf

## Test Plan

n/a

### DIFF
--- a/package.json
+++ b/package.json
@@ -72,8 +72,8 @@
     "@babel/preset-env": "^7.1.6",
     "@expo/webpack-config": "^0.16.0",
     "@react-native-community/eslint-config": "^3.0.0",
-    "@semantic-release/changelog": "^5.0.1",
-    "@semantic-release/git": "9.0.0",
+    "@semantic-release/changelog": "^6.0.0",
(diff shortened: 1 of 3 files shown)


package.json
```diff
@@ -91,9 +91,12 @@
     "react-native-web": "^0.17.0",
     "react-native-windows": "^0.64.0",
     "react-test-renderer": "17.0.1",
-    "semantic-release": "^17.4.6",
+    "semantic-release": "^19.0.0",
     "typescript": "^4.5.0"
   },
+  "resolutions": {
+    "npm/chalk": "^4.1.2"
```
krizzu, line 98: is it dependency of semantic-release?

krizzu (approved):
👍

package.json
tido64: I had this issue in another repo where the latest version of semantic-release bumped npm, which has a dependency on `chalk@*`, and because chalk@5 is ESM only, the pipeline failed because npm can't import it.

PR #760, opened 2022-02-26
fix(useAsyncStorage): make more robust

pke:
## Summary

Fixes the broken `useAsyncStorage` and introduces a new `useAsyncStorageValue` as a true hook replacement.

I think we should not change the signature of useAsyncStorage for now in a patch release.
In this patch release we could also already introduce the new `useAsyncStorageValue` which will be renamed to `useAsyncStorage` in the next major version.

I'll also add `watch` support to make the hook truly magic and work across multiple components.
I'll first add the watch support to the hooks infrastructure only, that means any changes done by AsyncStorage.setXXX/mergeXXX are not detected. But when one only uses the hooks, then changes in values are propagated into each hook.

## Test Plan

`yarn jest __tests___` for now.

### DIFF
--- a/__tests__/useAsyncStorage.test.ts
+++ b/__tests__/useAsyncStorage.test.ts
@@ -1,0 +1,84 @@
+/**
+ * @format
+ */
+/* eslint-disable no-shadow */
+import { renderHook, act } from '@testing-library/react-hooks';
+
+import AsyncStorage, { useAsyncStorage } from '../src';
+
+afterEach(AsyncStorage.clear);
+
+function expectStableCallbacks(result: any, rerender: () => void) {
+  const previousFunctions = Object.values(result.current).filter(
+    (item) => typeof item === 'function'
+  );
+  rerender();
+  const nowFunctions = Object.values(result.current).filter(
+    (item) => typeof item === 'function'
+  );
+  expect(nowFunctions).toHaveLength(previousFunctions.length);
+  expect(nowFunctions).toEqual(previousFunctions);
+}
+
+describe('useAsyncStorage', () => {
+  it('should export only stable output', async () => {
+    const { result, rerender } = renderHook(() => useAsyncStorage('key'));
+    expect(result.error).toBeUndefined();
+    expectStableCallbacks(result, rerender);
+  });
+
+  it('can read/write data to/from storage', async () => {
+    const newData = Math.floor(Math.random() * 1000).toString();
+    const { result } = renderHook(() => useAsyncStorage('key'));
+
+    await result.current.setItem(newData);
+
+    const data = await result.current.getItem();
+
+    expect(data).toBe(newData);
+  });
+
+  it('can remove from storage', async () => {
+    await AsyncStorage.setItem('key', 'value');
+    const { result } = renderHook(() => useAsyncStorage('key'));
+    await act(() => result.current.removeItem());
+    expect(await AsyncStorage.getItem('key')).toBeNull();
+  });
+
+  it.skip('should throw when mergeItem is not supported', async () => {
+    const { result } = renderHook(() => useAsyncStorage('key'));
+    const mergeItem = AsyncStorage.mergeItem;
+    delete AsyncStorage.mergeItem;
+    expect(result.current.mergeItem({})).rejects.toThrow();
+    AsyncStorage.mergeItem = mergeItem;
+  });
+
+  it('can use merge with current data in storage', async () => {
+    let originalPerson = {
+      name: 'Jerry',
+      age: 21,
+      characteristics: {
+        hair: 'black',
+        eyes: 'green',
+      },
+    };
+
+    const { result } = renderHook(() => useAsyncStorage('person'));
+
+    await result.current.setItem(JSON.stringify(originalPerson));
+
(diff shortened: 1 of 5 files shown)


pke:
The current jest tests in the example folder don't seem to work. 

I'd suggest to treat the example app as its own project with own package file instead.

The libs dependency should not carry an dependencies/tests the example project needs.

src/hooks.ts
```diff
@@ -1,13 +1,75 @@
...
+
+  return {
+    getItem,
+    setItem,
+    mergeItem,
+    removeItem,
   };
 }
+
+export function useAsyncStorageValue(
```
krizzu, line 35: I'd rather avoid adding more hooks into the mix. Fixing `useAsyncStorage` + tests would be sufficient for now 🙏

pke: OK. I'll update the PR and remove the code.

krizzu:
Hey @pke, do you need help with anything?

package.json
```diff
@@ -74,6 +74,8 @@
     "@react-native-community/eslint-config": "^3.0.0",
     "@semantic-release/changelog": "^6.0.0",
     "@semantic-release/git": "^10.0.0",
+    "@testing-library/react-hooks": "^7.0.2",
+    "@types/jest": "^27.4.1",
```
tido64, line 78: Is 27.x necessary? We're using Jest 26.x still, and this is pulling in 27.x dependencies, potentially conflicting with existing tests.

```suggestion
    "@types/jest": "^26.0.24",
```

src/hooks.ts
```diff
@@ -1,13 +1,35 @@
...
-    setItem: (...args) => AsyncStorage.setItem(key, ...args),
-    mergeItem: (...args) =>
+export function useAsyncStorage(key: string) {
+  const getItem = React.useCallback(
+    (...args) => AsyncStorage.getItem(key, ...args),
+    [key]
+  );
+
+  const setItem = React.useCallback(
+    //@ts-ignore
```
tido64, line 11: We shouldn't add any `@ts-ignore`. What's the error here? Can we fix it?

src/hooks.ts
```diff
@@ -1,13 +1,35 @@
...
+    (...args) => AsyncStorage.setItem(key, ...args),
+    [key]
+  );
+
+  const mergeItem = React.useCallback(
+    (...args) =>
+      //@ts-ignore
```
tido64, line 18: We shouldn't add any `@ts-ignore`. What's the error here? Can we fix it?

src/hooks.ts
```diff
@@ -1,13 +1,35 @@
...
+export function useAsyncStorage(key: string) {
+  const getItem = React.useCallback(
+    (...args) => AsyncStorage.getItem(key, ...args),
+    [key]
+  );
+
+  const setItem = React.useCallback(
+    //@ts-ignore
+    (...args) => AsyncStorage.setItem(key, ...args),
+    [key]
```
tido64, line 13: Do we have to create a new array for each of the callbacks? Can we create it once and reuse?

yarn.lock
```diff
@@ -1098,6 +1098,13 @@
   dependencies:
     regenerator-runtime "^0.13.4"
 
+"@babel/runtime@^7.12.5":
```
tido64, line 1101: Can you run `npx yarn-deduplicate` in the repo to remove some of these duplicates?

__tests__/useAsyncStorage.test.ts
```diff
@@ -0,0 +1,84 @@
+/**
+ * @format
+ */
+/* eslint-disable no-shadow */
```
tido64, line 4: Why is this ESLint disable necessary?

tido64 (commented):
Thanks for fixing the hooks! I've left some comments.

pke:
After doing some more jest testing in my own project I came to the conclusion that the RNAS tests are somewhat broken. Maybe the mock is broken. I need to investigate more.

src/hooks.ts
pke: Yes its required because TS complains:
[ts] src/hooks.ts(11,44): error TS2556: A spread argument must either have a tuple type or be passed to a rest parameter.

src/hooks.ts
pke: No, this doesn't work. Eslint will complain:
```tsx
const dep = [key]
const getItem = React.useCallback(
  (...args) => AsyncStorage.getItem(key, ...args),
  dep
);
```

```
const dep: string[]

React Hook React.useCallback was passed a dependency list that is not an array literal. This means we can't statically verify whether you've passed the correct dependencies.eslint[react-hooks/exhaustive-deps](https://github.com/facebook/react/issues/14920)
React Hook React.useCallback has a missing dependency: 'key'. Either include it or remove the dependency array.eslint[react-hooks/exhaustive-deps](https://github.com/facebook/react/issues/14920)
```

pke:
I have not much time to work on that further, since I also moved to an alternative sync lib for settings.

The tests for the hooks don't seem to run, not sure whats wrong with the setup there. Complaining about not able to find the RNAS mock.

Issue #764, opened 2022-03-08
multiGet return type incorrect?


mtt87:
### What happened?

👋  while using the library I noticed that `multiGet` has this signature:
```ts
multiGet: (keys: string[], callback?: MultiGetCallback) => Promise<readonly KeyValuePair[] | void>;
```
I'm not sure the return type is correct
```ts
Promise<readonly KeyValuePair[] | void>
```
It seems like `void` doesn't make sense, if I understood correctly how it works when a value doesn't exist it will return `[string null]` so the `KeyValuePair[]` is always the return type, not sure in which scenario it would return `void`.

I noticed it because Typescript is not happy when doing this
```ts
const values = await AsyncStorage.multiGet([ KEY1, KEY2, KEY3 ]);
setSomething(values[0][1]);
```
Showing this type error
```
Element implicitly has an 'any' type because expression of type '0' can't be used to index type 'void | readonly KeyValuePair[]'.
  Property '0' does not exist on type 'void | readonly KeyValuePair[]'.
```

### Version

1.16.1

### What platforms are you seeing this issue on?

- [ ] Android
- [ ] iOS
- [ ] macOS
- [ ] Windows
- [ ] web

### System Information

```true
n/a
```

### Steps to Reproduce

n/a

tido64:
I agree it's a bit weird. `void` doesn't really exist in JS as it would simply return `undefined`, or `null` in this case. In TypeScript, it's fine to use `void` for a function that may not return anything. The fix to your example is to add a check:

```ts
const values = await AsyncStorage.multiGet([ KEY1, KEY2, KEY3 ]);
if (values) {
  setSomething(values[0][1]);
}
```

This will allow TypeScript to narrow the type to `KeyValuePair[]` (since `values` cannot be `void` inside the if-statement), and you'll be able to index is as normal.

mtt87:
@tido64 thanks for your reply

> In TypeScript, it's fine to use `void` for a function that may not return anything. 

Sure but is this really the case? Can this function not return anything?
It's 2 different things to not return anything (`void` or `undefined` or `null`) or return an array of `KeyValuePair` that has `[string, null]` when the key is not in the storage or `[string, string]` when the storage has the key.

What I'm saying here is that the library should probably change
```ts
Promise<readonly KeyValuePair[] | void>
```
to
```ts
Promise<readonly KeyValuePair[]>
```

mtt87:
[...]

PR #767, opened 2022-03-09
Update types.ts to fix wrong void return type of multiGet

mtt87:
Fix https://github.com/react-native-async-storage/async-storage/issues/764

### DIFF
--- a/src/types.ts
+++ b/src/types.ts
@@ -23,7 +23,7 @@ export type MultiRequest = {
   keys: readonly string[];
   callback?: MultiGetCallback;
   keyIndex: number;
-  resolve?: (result?: readonly KeyValuePair[]) => void;
+  resolve?: (result: readonly KeyValuePair[]) => void;
   reject?: (error?: any) => void;
 };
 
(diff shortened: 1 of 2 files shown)


src/types.ts
```diff
@@ -125,7 +125,7 @@ export type AsyncStorageStatic = {
   multiGet: (
     keys: string[],
     callback?: MultiGetCallback
-  ) => Promise<readonly KeyValuePair[] | void>;
+  ) => Promise<readonly KeyValuePair[]>;
```
krizzu, line 128: you'll also need to update the type of [returned promise here](https://github.com/react-native-async-storage/async-storage/blob/master/src/AsyncStorage.native.ts#L272)

mtt87: Thank you I missed it 😄

krizzu:
There's one last bit that needs to be one - [updating the type of MultiRequest type](https://github.com/react-native-async-storage/async-storage/blob/master/src/types.ts#L26).

```diff
-  resolve?: (result?: readonly KeyValuePair[]) => void;
+  resolve?: (result: readonly KeyValuePair[]) => void;
```

tido64 (approved):

tido64:
Sorry I couldn't find time for this. I'm glad that you two managed to figure this out 😄

krizzu (approved):
Great 👏

PR #788, opened 2022-03-30
fix: cannot find `react-native-test-app` when Metro starts

tido64:
## Summary

Consumers are getting "Cannot find module 'react-native-test-app'" when starting Metro dev server.

## Test Plan

n/a

### DIFF
--- a/react-native.config.js
+++ b/react-native.config.js
@@ -1,10 +1,32 @@
-const fs = require('fs');
-const path = require('path');
-const {
-  androidManifestPath,
-  iosProjectPath,
-  windowsProjectPath,
-} = require('react-native-test-app');
+const project = (() => {
+  const fs = require('fs');
+  const path = require('path');
+  try {
+    const {
+      androidManifestPath,
+      iosProjectPath,
+      windowsProjectPath,
+    } = require('react-native-test-app');
+    return {
+      android: {
+        sourceDir: path.join('example', 'android'),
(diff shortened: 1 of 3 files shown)


react-native.config.js
```diff
@@ -1,10 +1,32 @@
...
+        project: iosProjectPath('example/ios'),
+      },
+      windows: fs.existsSync('example/windows/AsyncStorageExample.sln') && {
+        sourceDir: path.join('example', 'windows'),
+        solutionFile: 'AsyncStorageExample.sln',
+        project: windowsProjectPath(path.join(__dirname, 'example', 'windows')),
+      },
+    };
+  } catch (_) {
+    return undefined;
```
krizzu, line 27: so what's exactly issue here? why config might throw?

tido64: If you're consuming AsyncStorage, cli imports `react-native.config.js` and can't require `react-native-test-app` because it's a dev dependency.

krizzu: aa, got it 👍 
I'd rename `project` to maybe `exampleProjectSetup`, but that's super nitty, so ignore me 😄

krizzu (approved):

react-native.config.js
tido64: I'm going to keep it since it means we can shorthand the object below, and because you didn't feel strongly about it. :smile:

PR #803, opened 2022-05-09
fix: `multiGet` accepts read-only array

maxhungry:
## Summary

Since `getAllKeys` returns `readonly string[]`, it make sense for other functions to accept keys in `readonly`.

Error screenshot:

Screen Shot 2022-05-09 at 9 55 37 PM

## Test Plan

TS compiles with no errors; tests are passing.


tido64 (approved):
Thanks for fixing this ❤️

PR #817, opened 2022-06-23
fix: declare support for react-native 0.69

tido64:
## Summary

AsyncStorage should still work fine with react-native 0.69

## Test Plan

n/a

### DIFF
--- a/package.json
+++ b/package.json
@@ -65,7 +65,7 @@
     "merge-options": "^3.0.4"
   },
   "peerDependencies": {
-    "react-native": "^0.0.0-0 || 0.60 - 0.68 || 1000.0.0"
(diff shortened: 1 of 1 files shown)


package.json
```diff
@@ -65,7 +65,7 @@
     "merge-options": "^3.0.4"
   },
   "peerDependencies": {
-    "react-native": "^0.0.0-0 || 0.60 - 0.68 || 1000.0.0"
+    "react-native": "^0.0.0-0 || 0.60 - 0.69 || 1000.0.0"
```
krizzu, line 68: Why not we just go ahead and support any version 60+?

tido64: At some point native modules will be deprecated, and this is a really good way to get people to upgrade to the latest version 😛

krizzu: make sense!

krizzu (approved):

PR #823, opened 2022-07-21
chore: bump react-native to 0.68

tido64:
## Summary

Bumps react-native to 0.68.

## Test Plan

Make sure that `yarn start:*` works:

- [x] Android
- [x] iOS
- [x] macOS
- [x] Windows


tido64:
@krizzu: Would you mind taking a look at Android? I got build errors I'm not sure how to proceed on. I'll take a look on Windows in the meantime.

krizzu:
@tido64 so the problem is that test-app has bumped to Kotlin 1.7. So we have to solutions:

- Use KSP instead of KAPT for annotation processing. This requires few steps, [like described in this SO post](https://stackoverflow.com/a/72978960).
- Update Room to 2.5.0-alpha02, which has Kotlin 1.7 compatibility 

With first one, it's hard to assess how much of impact it would be on project that'd depend on AsyncStorage, which are still on lower version of Kotlin. It'd be good to test this KSP integration with lower Kt version to see if it's still compatible. 

Second one has drawback of Room requiring compilation/target sdk of 32. That means developer should migrate in project to those version and check compatibility of the rest of the app.

I'd be in favor on going with first approach, but would have to test the scenario first.

tido64:
> @tido64 so the problem is that test-app has bumped to Kotlin 1.7

I've added a flag to the test app so we can set our own Kotlin version and opened #832 to track support of 1.7. Everything should be building again now.

krizzu (approved):
Yes, let's not block it 👌

PR #828, opened 2022-08-04
build(ci): Improve Android e2e

krizzu:
## Summary

Bumping few bits in our CircleCI config. Using Linux machine instead of macos to run Android e2e. Using Test Butler to make sure e2e goes smoothly.

## Test Plan

Green CI

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -18,29 +18,28 @@ checkout step for each job: &addWorkspace
 default config for js: &js_defaults
   <<: *defaults
   docker:
-    - image: circleci/node:14
+    - image: cimg/node:14.18.1
 
 default config for macOS: &macos_defaults
   <<: *defaults
   resource_class: 'medium'
   macos:
-    xcode: '13.2.1'
+    xcode: '13.1.0'
 
-config for macOS (android): &macos_defaults_android
+config for android e2e: &android_e2e
   <<: *defaults
-  resource_class: 'medium'
-  macos:
-    xcode: '11.5.0'
+  resource_class: large
+  machine:
+    image: android:2022.06.2
 
 default config for android apk builds: &android_defaults
   <<: *defaults
   docker:
-    - image: reactnativecommunity/react-native-android
-  resource_class: 'medium'
-  working_directory: ~/async_storage
+    - image: cimg/android:2021.10.2-node # node 14.18.1
+  resource_class: medium
   environment:
-    - _JAVA_OPTIONS: '-XX:+UnlockExperimentalVMOptions -Xmx2048m'
-    - BUILD_THREADS: 2
+    JAVA_OPTIONS: '-XX:+UnlockExperimentalVMOptions -Xmx2048m'
+    BUILD_THREADS: 2
 
 # ==============================
 # CACHE CONFIG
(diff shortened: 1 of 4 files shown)


.circleci/config.yml
```diff
@@ -18,37 +18,35 @@ checkout step for each job: &addWorkspace
...
   docker:
-    - image: circleci/node:14
+    - image: cimg/node:14.18.1
 
 default config for macOS: &macos_defaults
   <<: *defaults
   resource_class: 'medium'
   macos:
-    xcode: '13.2.1'
+    xcode: '13.1.0'
```
tido64, line 27: Why are we downgrading Xcode here?

.circleci/config.yml
```diff
@@ -18,37 +18,35 @@ checkout step for each job: &addWorkspace
...
+  machine:
+    image: android:2022.06.2
 
 default config for android apk builds: &android_defaults
   <<: *defaults
   docker:
-    - image: reactnativecommunity/react-native-android
-  resource_class: 'medium'
-  working_directory: ~/async_storage
+    - image: cimg/android:2021.10.2-node # node 14.18.1
```
tido64, line 38: Shouldn't this also use `cimg/node:14.18.1` as above? Or `android:2022.06.2`?

.circleci/config.yml
```diff
@@ -229,71 +216,48 @@ jobs:
...
-            echo 'export ANDROID_HOME="/usr/local/share/android-sdk"' >> $BASH_ENV
-            echo 'export ANDROID_SDK_ROOT="/usr/local/share/android-sdk"' >> $BASH_ENV
-            echo 'export PATH="$ANDROID_SDK_ROOT/emulator:$ANDROID_SDK_ROOT/tools:$ANDROID_SDK_ROOT/platform-tools:$PATH"' >> $BASH_ENV
-            echo 'export QEMU_AUDIO_DRV=none' >> $BASH_ENV
-            echo 'export JAVA_HOME=$(/usr/libexec/java_home)' >> $BASH_ENV
-            source $BASH_ENV
-      - restore-cache: *cache_restore_brew_android
+            curl -o- https://raw.githubusercontent.com/nvm-sh/nvm/v0.38.0/install.sh | bash
+            source ~/.bashrc
+            nvm install v14.18.1
```
tido64, line 227: Why do we install Node here when we specify the Node version above?

.circleci/config.yml
krizzu: Yeah, I went one minor down, as 13.1.0 has node `14.18.1`, while 13.2.1 has `16.13.1`, to keep version consistent across executors

.circleci/config.yml
krizzu: This image has Android SDK preinstalled, and is used to create an apk for tests. I could merge the apk creation into e2e job, since it's no longer macos.

tido64: Is there a cost to spinning up a separate machine? I'd imagine we should stay on the same machine for as long as possible, but I don't know how CircleCI works.

krizzu: I think there's no cost for boot up, but rather the overall execution time.

tido64 (approved):
I'd like for CI to use less machines to make the config easier to maintain. But if we intend to migrate to GitHub Actions anyway, maybe it's fine for now.

krizzu:
> I'd like for CI to use less machines to make the config easier to maintain. But if we intend to migrate to GitHub Actions anyway, maybe it's fine for now.

I changed the config to use one executor for Android, that should simplify this a bit 👍

tido64 (approved):

PR #841, opened 2022-09-06
ci: set up ccache for macOS builds

tido64:
## Summary

Set up ccache for macOS builds.

## Test Plan

CI should pass. First builds should take as much time as it does today, but subsequent ones should be significantly faster.

| Before | After |
| - | - |
| [24m 32s](https://github.com/react-native-async-storage/async-storage/runs/8203935502?check_suite_focus=true) | [13m 31s](https://github.com/react-native-async-storage/async-storage/runs/8204398679?check_suite_focus=true) |

### DIFF
--- a/.ccache/ccache.conf
+++ b/.ccache/ccache.conf
@@ -1,0 +1,5 @@
+depend_mode = true
+direct_mode = true
+hard_link = true
+run_second_cpp = true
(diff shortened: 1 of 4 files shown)


krizzu (approved):
👏

PR #846, opened 2022-10-03
Added fallback for web when localStorage is not available

kurtsergey:


krizzu:
Hey @kurtsergey 

I appreciate the PR, but I don't see any prior discussion about it, so I have no context. Can we close it for now and discuss what you try to achieve?

PR #870, opened 2022-11-11
fix: declare support for react-native 0.71

varemenos:
## Summary

Adds support for RN 0.71


tido64 (approved):
Thanks for doing this :heart:

PR #894, opened 2023-01-09
ci: bump GitHub actions

tido64:
## Summary

We're seeing deprecation warnings on CI such as here: https://github.com/react-native-async-storage/async-storage/actions/runs/3864459872

Bumping all the actions to their latest versions to hopefully get rid of them.

## Test Plan

n/a


krizzu (approved):
👍

PR #910, opened 2023-02-06
Convert library to TurboModule

j-piasecki:
## Summary

This PR adds `FabricExample` app with the new arch enabled and converts the library to TurboModule.

Notable changes:
- renamed Android module from `RNC_AsyncSQLiteDBStorage` to `RNCAsyncStorage`, matching iOS name
- split the Package file into two: one for the Java implementation, one for the Kotlin one. This is mainly because of `@ReactModuleList` annotation. Previously reflection and `BuildConfig` field were used to check if the Kotlin module should be used, but the annotation expects static declaration. I've achieved this using gradle source sets, to conditionally include one of the modules.
- the generated spec is committed to the repository and used on the old architecture, which simplifies the inheritance. In case the spec changes, this file would need to be upgraded 

## Test Plan

Check the example app. On Android the new arch is enabled by default, while on iOS the `RCT_NEW_ARCH_ENABLED=1` flag is required when installing pods.

PR #911, opened 2023-02-07
Add `FabricExample` application with the new arch enabled

j-piasecki:
## Summary

This PR adds a new example app using RN 0.71 with the new architecture enabled by default. This was used to test https://github.com/react-native-async-storage/async-storage/pull/910.

It's not necessary for this PR to be merged in order to use the one converting the library to turbomodule as the app may be removed. Setting the base to this one should make relevant changes easier to check.


tido64:
Hi, thanks for working on this. I see you're trying to add TurboModule support. Have you had a look at how the other example app works? We are using [react-native-test-app](https://github.com/microsoft/react-native-test-app), which allows us to easily enable TurboModule (for both [Android](https://github.com/microsoft/react-native-test-app/wiki/Android-Specifics#fabric-and-turbomodule) and [iOS/macOS](https://github.com/microsoft/react-native-test-app/wiki/iOS-and-macOS-Specifics#fabric-and-turbomodule)). Would it make more sense to reuse the existing test app and instead work on enabling building of both Legacy and New Architecture on CI?

cc @krizzu

j-piasecki:
I don't have experience with `react-native-test-app`, what's the correct way to upgrade RN version? I didn't see anything about it on the GitHub wiki.

A potential problem could be that the library would work only on the RN 0.70 and higher when using the new arch due to a different codegen config, and It seems like [neither 0.70, nor 0.71](https://github.com/microsoft/react-native-test-app/wiki#react-native-versions) is fully supported yet.

I could close this PR and remove the commits related to the new app from the other PR so you could set it up correctly.

j-piasecki:
I've tried to implement it, but after upgrading to React Native 0.71, the `pod install` started failing with:
```
[!] Invalid `Podfile` file: 783: unexpected token at 'Config Validation Error: "project.ios.project" is not allowed
```
I've changed the [`project`](https://github.com/react-native-async-storage/async-storage/blob/a9e9f48e8635e58618f1a1ddb4245c2fa68ea7f8/react-native.config.js#L18) field to `sourceDir` and got the following:
```
Invalid `Podfile` file: undefined method `[]' for nil:NilClass.

 #  from /Users/jakubpiasecki/Projects/react-native-async-storage/example/ios/Podfile:6
 #  -------------------------------------------
 #  use_flipper! false
 >  use_test_app! do |target|
 #    target.app do
 #  -------------------------------------------
```


On Android I had to add `@OptIn(ExperimentalStdlibApi::class)` annotation to [this](https://github.com/facebook/react-native/blob/cab865be797b724d2fda5441e0ef23559180f722/packages/react-native-gradle-plugin/src/main/kotlin/com/facebook/react/tasks/BundleHermesCTask.kt#L136) method due to `buildList` being marked with `ExperimentalStdlibApi`.

After that I got:
```
Execution failed for task ':react-native-async-storage_async-storage:buildCodegenCLI'.
> A problem occurred starting process 'command '/Users/jakubpiasecki/Projects/react-native-async-storage/example/node_modules/react-native-codegen/scripts/oss/build.sh''
```
which I believe is due to https://github.com/facebook/react-native/issues/35495.

tido64:
> I've tried to implement it, but after upgrading to React Native 0.71, the `pod install` started failing with…

https://github.com/react-native-async-storage/async-storage/pull/915 should fix it.

> On Android I had to add `@OptIn(ExperimentalStdlibApi::class)` annotation to [this](https://github.com/facebook/react-native/blob/cab865be797b724d2fda5441e0ef23559180f722/packages/react-native-gradle-plugin/src/main/kotlin/com/facebook/react/tasks/BundleHermesCTask.kt#L136) method due to `buildList` being marked with `ExperimentalStdlibApi`.

This one I'm unfamiliar with. @krizzu do you know why we have to do this?

krizzu:
> > On Android I had to add `@OptIn(ExperimentalStdlibApi::class)` annotation to [this](https://github.com/facebook/react-native/blob/cab865be797b724d2fda5441e0ef23559180f722/packages/react-native-gradle-plugin/src/main/kotlin/com/facebook/react/tasks/BundleHermesCTask.kt#L136) method due to `buildList` being marked with `ExperimentalStdlibApi`.
> 
> This one I'm unfamiliar with. @krizzu do you know why we have to do this?

What version of Kotlin is used? In 1.6 [buildList became stable](https://kotlinlang.org/docs/whatsnew16.html#stable-collection-builders)

j-piasecki:
@krizzu You're right, it seems like my Android Studio decided to use Kotlin 1.5. I've just tried building from CLI and I didn't get this error.

j-piasecki:
> https://github.com/react-native-async-storage/async-storage/pull/915 should fix it.

It fixes the first problem, however the second one:
```
Invalid `Podfile` file: undefined method `[]' for nil:NilClass.

 #  from /Users/jakubpiasecki/Projects/react-native-async-storage/example/ios/Podfile:6
 #  -------------------------------------------
 #  use_flipper! false
 >  use_test_app! do |target|
 #    target.app do
 #  -------------------------------------------
```
is still there.

tido64:
> is still there.

Sorry, I missed that CLI requires `sourceDir`. I've updated the PR and `pod install` _should_ work now. Please make sure you run `pod install` from `example/ios` and not use `--project-directory=example/ios`. I've just discovered that a bug that I thought I had fixed in `react-native` related to this.

j-piasecki:
Ok, I've got it to work on both Android and iOS. For now I've added a quick fix for the codegen issue linked above: https://github.com/react-native-async-storage/async-storage/pull/910/commits/abed02f979994e36f36baf57b7be36d7016f8d28#diff-ed8534ece8080a880075e103269eb1b9bdff22e8dfcf8427d7d977603292b889R59-R65. I guess this PR can now be closed  as https://github.com/react-native-async-storage/async-storage/pull/910 no longer requires it?

Thanks for your help!

tido64:
> Ok, I've got it to work on both Android and iOS.

Thank you for working on this. Please let me know if you hit other issues with the test app. I'll try to wrap up #915 soon.

Issue #905, opened 2023-01-30
Include information on the storage implemention


jackbridger:
### Proposal

Thanks for your hard work on the project.
I was curious what the implementation looks like under the hood in Android and iOS and I couldn't find the info on the readme or docs (I might just have missed it).
I think this might be a nice addition to help us understand at a glance how this great package works. 
If this isn't helpful please feel free to close!

### Implementation Details

It could include something like the below:

Where your Redux store is stored:
* Android - SQLite
* iOS - small values in serialized dictionary and larger values in separate files
* Web - window.localStorage
* MacOS - I couldn't find the details
* Windows - I couldn't find the details

krizzu:
I like the idea - we could add a section in the documentation about it.

And to fill up some gaps:

```diff
 Where your Redux store is stored:
 
Android - SQLite
+ iOS - small values (not exceeding 1024 characters) are serialized and stored in a common `manifest.json` file, while larger values are stored in individual, dedicated files (named as MD5 hashed `key`)
Web - window.localStorage
+ MacOS - Same as iOS
+ Windows - SQLite
```

jackbridger:
Great, glad you like it!

I can make a PR for this early next week

PR #912, opened 2023-02-07
Add info on where data is stored

jackbridger:
## Summary

This is a docs change and relates to issue #905  - adding information on where data is stored.

I placed it under the advanced category in the sidebar. Let me know if you prefer it somewhere else. 

CleanShot 2023-02-07 at 10 49 59

## Test Plan

It is a docs change so likely just need to review that the docs run as they should. Let me know if you want me to do any other testing.

Thanks!!

### DIFF
--- a/website/docs/advanced/Where-data-stored.md
+++ b/website/docs/advanced/Where-data-stored.md
@@ -4,8 +4,8 @@ title: Where your data is stored
 sidebar_label: Where data is stored
 ---
 
-* <b>Android</b> - SQLite
-* <b>iOS</b> - small values (not exceeding 1024 characters) are serialized and stored in a common `manifest.json` file, while larger values are stored in individual, dedicated files (named as MD5 hashed `key`)
-* <b>Web</b> - window.localStorage
+* **Android** - SQLite
+* **iOS** - small values (not exceeding 1024 characters) are serialized and stored in a common `manifest.json` file, while larger values are stored in individual, dedicated files (named as MD5 hashed `key`)
 * **macOS** - Same as iOS
-* <b>Windows</b> - SQLite
+* **Web** - window.localStorage
+* **Windows** - SQLite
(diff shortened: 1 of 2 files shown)


website/docs/advanced/Where-data-stored.md
```diff
@@ -0,0 +1,11 @@
+---
+id: where_data_stored
+title: Where your data is stored
+sidebar_label: Where data is stored
+---
+
+* <b>Android</b> - SQLite
```
tido64, line 7: Can we replace `<b>` and `</b>` with `**` instead?
```suggestion
* **Android** - SQLite
```

website/docs/advanced/Where-data-stored.md
```diff
@@ -0,0 +1,11 @@
...
+* <b>iOS</b> - small values (not exceeding 1024 characters) are serialized and stored in a common `manifest.json` file, while larger values are stored in individual, dedicated files (named as MD5 hashed `key`)
+* <b>Web</b> - window.localStorage
+* <b>MacOS</b> - Same as iOS
```
tido64, line 10: _macOS_ is stylized with a lower case "m"

```suggestion
* **macOS** - Same as iOS
```

website/docs/advanced/Where-data-stored.md
```diff
@@ -0,0 +1,11 @@
...
+* <b>Windows</b> - SQLite
```
tido64, line 11: Nit: It would also be nice if we could sort these alphabetically.

tido64 (commented):
Thanks for adding this 👍

website/docs/advanced/Where-data-stored.md
jackbridger: good point!

website/docs/advanced/Where-data-stored.md
jackbridger: Updating now!

jackbridger:
Everything should be updated now
CleanShot 2023-02-07 at 11 21 29

tido64 (approved):
Thanks 🙏

krizzu:
Thanks Jack 🙏 
@tido64 seems like Windows cannot resolve some dependencies, again. I think we can just move this forward anyway.

krizzu (approved):

jackbridger:
Yay! Thanks both for your support. This was my first contribution to a project that lots of people use. Even though it's only docs, it feels good. Hope it's the start of more contributions. Thanks!!

PR #914, opened 2023-02-07
chore: bump Yarn to 3.4.1

tido64:
## Summary

We keep getting network failures when running `yarn install`. Yarn v3 handles this much better.

## Test Plan

CI should pass.


package.json
```diff
@@ -37,7 +37,6 @@
     "url": "https://github.com/react-native-async-storage/async-storage.git"
   },
   "scripts": {
-    "ci": "yarn --pure-lockfile --non-interactive",
```
tido64, line 40: Note that this is default in v3, when it detects that it is running in a CI environment.

krizzu (approved):

PR #920, opened 2023-02-20
ci: move JS jobs from CircleCI to GH

tido64:
## Summary

Move JS related jobs from CircleCI to GitHub Actions.

## Test Plan

CI should pass.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -5,6 +5,26 @@ on:
       - master
   pull_request:
 jobs:
+  review:
+    name: 'Review'
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v3
+      - name: Set up Node.js
+        uses: actions/setup-node@v3.6.0
+        with:
+          node-version: 16
+          cache: 'yarn'
+      - name: Install JS dependencies
+        run: |
+          yarn
+      - name: Lint
+        run: |
+          yarn test:lint
+      - name: TypeScript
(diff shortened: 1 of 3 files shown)


.github/workflows/ci.yml
```diff
@@ -5,6 +5,26 @@ on:
...
+          cache: 'yarn'
+      - name: Install JS dependencies
+        run: |
+          yarn
+      - name: Lint
+        run: |
+          yarn test:lint
+      - name: TypeScript
+        run: |
+          yarn test:ts
```
krizzu, line 27: What about adding a cache step to speed things up in future?
```
      - name: Cache save/restoration
        uses: actions/cache@v3
        id: cache-deps
        with:
          path: node_modules
          key: ${{ runner.os }}-yarn-v1-${{ hashFiles('yarn.lock') }}
      - name: Install dependencies
        if: steps.cache-deps.outputs.cache-hit != 'true'
        run: yarn
```

tido64: `setup-node` caches `.yarn/cache`. If this is populated, `yarn install` takes only a few seconds. This cache is smaller, and we don't get stale cache (e.g. Babel, Metro, etc.) that lives in `node_modules`.

krizzu: Make sense! On top of that:
> Note It is not recommended to cache node_modules, as it can break across Node versions and won't work with npm ci
TIL 😄

krizzu (approved):
💪

PR #932, opened 2023-03-07
refactor: introduce `AsyncStorageFactory`

tido64:
## Summary

Introduces `AsyncStorageFactory` for v2.

## Test Plan

image


tido64:
@krizzu Should this go in before we make `main` default? Or did you want to approach this a different way?

krizzu:
@tido64 let's come back to this once moved to main and monorepo. We can always cherry pick it

PR #936, opened 2023-03-13
ci: fix website deployment

tido64:
## Summary

Looks like website deployment broke: https://github.com/react-native-async-storage/async-storage/actions/runs/4402619064/jobs/7709976697

## Test Plan

`yarn.lock` has been updated. Not entirely sure if there is anything else to test.


krizzu (approved):
Thanks 👍

PR #941, opened 2023-03-13
refactor: add TurboModule spec

tido64:
## Summary

We don't have a TurboModule implementation yet, but this should be backwards compatible with legacy native module.

## Test Plan

n/a

### DIFF
--- a/src/RCTAsyncStorage.ts
+++ b/src/RCTAsyncStorage.ts
@@ -1,25 +1,66 @@
-// @ts-ignore Module '"react-native"' has no exported member 'TurboModuleRegistry'.
-import { NativeModules, TurboModuleRegistry } from 'react-native';
+import type { TurboModule } from 'react-native';
+import { TurboModuleRegistry } from 'react-native';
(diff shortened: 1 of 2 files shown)


src/RCTAsyncStorage.ts
```diff
@@ -1,25 +1,59 @@
...
-  // with undefined.
-  if (TurboModuleRegistry) {
-    RCTAsyncStorage =
-      TurboModuleRegistry.get('AsyncSQLiteDBStorage') ||
-      TurboModuleRegistry.get('AsyncLocalStorage');
-  } else {
-    RCTAsyncStorage =
-      NativeModules['AsyncSQLiteDBStorage'] ||
-      NativeModules['AsyncLocalStorage'];
+export interface Spec extends TurboModule {
```
tido64, line 6: @krizzu: `getAllKeys` and `multiMerge` are not specified in the v2 doc. Are we planning to deprecated them?

krizzu (approved):

PR #965, opened 2023-05-26
Allow consumption of RCTAsyncStorage as TurboModule

rozele:
## Summary

In react-native-windows, attributed C++ modules can be consumed as either a legacy native module or TurboModule. This change enables registration of RNCAsyncStorage as a TurboModule.

## Test Plan

Tested this out in a react-native-windows C++ app, where instead of using the [ReactPackageProvider](https://github.com/react-native-async-storage/async-storage/blob/master/windows/code/ReactPackageProvider.idl), I register [RNCAsyncStorage.h](https://github.com/react-native-async-storage/async-storage/blob/master/windows/code/RNCAsyncStorage.h) directly as a TurboModule.

### DIFF
--- a/src/RCTAsyncStorage.ts
+++ b/src/RCTAsyncStorage.ts
@@ -2,15 +2,18 @@
 import { NativeModules, TurboModuleRegistry } from 'react-native';
 import { shouldFallbackToLegacyNativeModule } from './shouldFallbackToLegacyNativeModule';
 
-let RCTAsyncStorage =
-  NativeModules['PlatformLocalStorage'] || // Support for external modules, like react-native-windows
-  NativeModules['RNC_AsyncSQLiteDBStorage'] ||
-  NativeModules['RNCAsyncStorage'];
+// TurboModuleRegistry falls back to NativeModules so we don't have to try go
+// assign NativeModules' counterparts if TurboModuleRegistry would resolve
+// with undefined.
+let RCTAsyncStorage = TurboModuleRegistry
+  ? TurboModuleRegistry.get('PlatformLocalStorage') || // Support for external modules, like react-native-windows
+    TurboModuleRegistry.get('RNC_AsyncSQLiteDBStorage') ||
+    TurboModuleRegistry.get('RNCAsyncStorage');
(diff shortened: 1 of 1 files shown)


src/RCTAsyncStorage.ts
```diff
@@ -2,15 +2,18 @@
...
 import { shouldFallbackToLegacyNativeModule } from './shouldFallbackToLegacyNativeModule';
 
-let RCTAsyncStorage =
-  NativeModules['PlatformLocalStorage'] || // Support for external modules, like react-native-windows
-  NativeModules['RNC_AsyncSQLiteDBStorage'] ||
-  NativeModules['RNCAsyncStorage'];
+// TurboModuleRegistry falls back to NativeModules so we don't have to try go
+// assign NativeModules' counterparts if TurboModuleRegistry would resolve
+// with undefined.
+let RCTAsyncStorage = TurboModuleRegistry
```
tido64, line 8: I wonder if we should just bump the minimum requirement to 0.61 so we don't have to do this check. What do you think @krizzu?

krizzu: I'm fine with the change, although that means we'd have to keep a compatibility table in readme (so that, we show which version is supported by which RN version). I guess it's safe to say we support 0.60 and up version

tido64: Okay, let's do this after this PR merges. Can you fix the Release job? It looks like it's unable to fetch user and password for some reason.

tido64 (approved):

src/RCTAsyncStorage.ts
krizzu: @tido64 sorry, missed the message
I created PR to separate release from Pull Request checks:
https://github.com/react-native-async-storage/async-storage/pull/972

The PRs from external contributor fails, because of nature of "pull_request" trigger preventing access to Secrets and limiting GH token to READ only

tido64: Ah, I didn't realize this was an issue. Thanks for looking into it. I guess that means we can merge this now?

krizzu: Yeah, let's get this in 🙏

krizzu (approved):

PR #972, opened 2023-06-06
ci: Separate release action from PR action

krizzu:
## Summary

~Removing release~ Making sure release action is run only on push actions, so that it would not fail, due to lack of access to secrets.

Also, running TS/Lint in a matrix as a bonus

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -8,6 +8,9 @@ jobs:
   review:
     name: Review
     runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        test-name: [lint, ts]
     steps:
       - name: Checkout
         uses: actions/checkout@v3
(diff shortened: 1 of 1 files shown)


.github/workflows/ci.yml
```diff
@@ -139,32 +138,3 @@ jobs:
       - name: Build
         run: |
           yarn react-native run-windows --release --arch x64 --logging --no-packager --no-launch --no-deploy --msbuildprops "BundleEntryFile=index.ts" --no-telemetry
-  release:
-    name: Release
-    needs: [review, android, ios, macos, windows]
-    runs-on: ubuntu-22.04
```
tido64, line 145: Wouldn't it be simpler to add `if: ${{ github.event_name == 'push' }}` here?

krizzu: aw yeah, it would, but the idea of separate workflows looked more appealing to me 😄 
Let me change it back

.github/workflows/ci.yml
```diff
@@ -17,14 +20,9 @@ jobs:
...
       - name: Install JS dependencies
-        run: |
-          yarn
-      - name: Lint
-        run: |
-          yarn test:lint
-      - name: TypeScript
-        run: |
-          yarn test:ts
+        run: yarn --frozen-lockfile
```
tido64, line 23: `--frozen-lockfile` is enabled by default when Yarn is run on CI.

krizzu: ahh, still holding to my v1 😄

PR #989, opened 2023-07-13
Update build.gradle

ghyath5:
Fixed react native >= 0.72 `Could not create an instance of type com.android.build.api.variant.impl.LibraryVariantBuilderImpl. Namespace not specified. Please specify a namespace in the module's build.gradle file like so`

### DIFF
--- a/android/build.gradle
+++ b/android/build.gradle
@@ -86,6 +86,7 @@ if (isNewArchitectureEnabled()) {
 }
 
 android {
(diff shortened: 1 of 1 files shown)


tido64:
Thank you! This should now be addressed in https://github.com/react-native-async-storage/async-storage/commit/ce19249ca5c2a9ea41464bb430f476b78d3223b4.

PR #1000, opened 2023-08-12
fix: tell users about how autolinking works

tido64:
## Summary

Tell users about how autolinking works.

## Test Plan

n/a


krizzu (approved):
Great stuff 🙏

PR #1002, opened 2023-08-14
chore: monorepo setup

krizzu:
## Summary

Moving to monorepo!
Scoping dependencies to each workspace (no hoist) as for now. In upcoming PRs will possibly move example to different workspace.


package.json
```diff
@@ -1,108 +1,9 @@
...
-    "android/build.gradle",
-    "android/src",
-    "android/testresults.gradle",
-    "ios/",
-    "jest/",
-    "lib/",
-    "macos/",
-    "src/",
-    "windows/"
+  "name": "@react-native-async-storage/project",
```
tido64, line 2: Some tools (like Nx) use the convention `@scope/root`:
```suggestion
  "name": "@react-native-async-storage/root",
```

packages/default-storage-backend/.clang-format
tido64, line 1: This should remain in root so that it gets applied to the whole repository.

packages/default-storage-backend/.gitignore
tido64, line 1: I wonder if this should remain in root as well. We should only have local `.gitignore` when necessary.

.github/workflows/ci.yml
```diff
@@ -168,3 +177,4 @@ jobs:
           git config user.email ${{ secrets.GH_BOT_EMAIL }}
           git config user.name ${{ secrets.GH_BOT_NAME }}
           yarn semantic-release
```
tido64, line 179: Note for later: We should look into replacing `semantic-release` with something that understands monorepos. In rnx-kit, [Changesets](https://github.com/changesets/changesets) has been working very well for us.

.github/workflows/ci.yml
```diff
@@ -23,6 +23,7 @@ jobs:
         run: yarn
       - name: Run test ${{ matrix.test-name }}
         run: yarn test:${{ matrix.test-name }}
+        working-directory: packages/default-storage-backend
```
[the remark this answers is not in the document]
krizzu: I thought more about TYPE_OF_STORAGE-storage-backend convention. We already have other storage called `sqlite-storage-backend`. Besides, I don't think we will have more storages in this monorepo anyway

package.json
krizzu: sounds good 👍

packages/default-storage-backend/.ccache/ccache.conf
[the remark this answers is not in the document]
krizzu: I saw ccache being referred only by macos script, so I updated the script to point to the new location

packages/default-storage-backend/.clang-format
krizzu: sounds good
How do you use this formatter anyway? Simply calling CLI?

packages/default-storage-backend/.gitignore
krizzu: It's fine by me to have one top-level gitignore to ignore common files, while per workspace (if required) to complement. I did some unification of those files now

.github/workflows/ci.yml
krizzu: Sweet, will remember that 🙏

packages/default-storage-backend/.clang-format
tido64: Yes, you should be able to run `clang-format` from anywhere. I usually do `clang-format -i $(git ls-files '*.cpp' '*.h' '*.m' '*.mm')` so that it doesn't pick up files from `node_modules` etc.

https://github.com/microsoft/react-native-test-app/blob/8e1ba826559915b21f9bcb730f8d920a830ec27c/package.json#L64C18-L64C76

.github/workflows/ci.yml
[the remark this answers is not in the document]
tido64: It would look a lot nicer in the directory if we stick to the storage-<category>-<implementation> convention though:
```
storage-backend-default   # What we have in 1.x
storage-backend-mmkv      # Maybe?
storage-backend-os-prefs  # NSUserDefaults (iOS) or SharedPreference (Android)
storage-backend-sqlite    # SQLite
storage-core              # JS
storage-types             # TypeScript interfaces
```
But I suppose we could also change this later, so I'll let you decide.

PR #1005, opened 2023-08-30
chore: post migration cleanup

krizzu:
## Summary

Moving files belonging to the storage workspace


packages/default-storage-backend/CHANGELOG.md
```diff
@@ -4,3 +4,207 @@
 ### Bug Fixes
```
krizzu, line 4: I moved the file from root (where changelog used to be) to the workspace

packages/default-storage-backend/package.json
```diff
@@ -60,6 +60,9 @@
     "test:e2e:ios": "scripts/ios_e2e.sh 'test'",
     "test:e2e:macos": "scripts/macos_e2e.sh 'test'"
   },
+  "installConfig": {
+    "hoistingLimits": "workspaces"
+  },
```
tido64, line 65: Do we still need this?

tido64 (approved):

packages/default-storage-backend/package.json
krizzu: yeah, without this, the deps will be hoisted which somehow breaks the example app. I opted out from having to set nohoist globally to this package only. Maybe we could move example app as separate package in future

PR #1006, opened 2023-08-30
chore: Scaffold Core repo

krizzu:
## Summary

A basic setup for Core repo. Typescript and Eslint config are in roots, extended by workspaces.


.eslintignore
```diff
@@ -0,0 +1 @@
+node_modules/
```
krizzu, line 1: Not really needed unless pointed by eslint to lint. Can be removed

.eslintrc
```diff
@@ -0,0 +1,6 @@
+{
+  "extends": ["@react-native-community"],
+  "rules": {
+    "no-console": 2
```
tido64, line 4: Can we use the string variant instead of a number? Easier to read:
```suggestion
    "no-console": "error"
```

.prettierrc
```diff
@@ -0,0 +1,7 @@
+{
+  "tabWidth": 2,
+  "semi": true,
+  "singleQuote": true,
+  "bracketSpacing": true,
+  "parser": "typescript"
+}
```
tido64, line 7: I'd love to remove this config at some point and just use whatever default Prettier uses.

packages/core/.gitignore
```diff
@@ -0,0 +1 @@
+lib
```
tido64, line 1: Can we put this into the top level `.gitignore`? `/packages/*/lib` should catch all?

packages/core/package.json
```diff
@@ -0,0 +1,44 @@
+{
+  "name": "@react-native-async-storage/core",
+  "version": "0.0.0",
+  "description": "Core API of Async Storage",
+  "main": "lib/commonjs/index",
+  "module": "lib/module/index",
+  "types": "lib/typescript/index.d.ts",
+  "react-native": "src/index",
+  "source": "src/index",
```
tido64, line 9: We should also include an exports map. Metro started supporting it recently.

Also, why are we missing the extensions? Platform forking is not applied to entry point files regardless.

.eslintignore
tido64: Let's remove it.

.eslintrc
tido64, line 1: Since this is only used as base, can we move it into a `.config` folder to hide the clutter in root?

tsconfig.base.json
tido64, line 1: Since this is only used as base, can we move it into a `.config` folder to hide the clutter in root?

packages/default-storage-backend/src/helpers.ts
```diff
@@ -1,3 +1,4 @@
+/* eslint-disable no-console */
```
tido64, line 1: Instead of disabling this rule for the whole file, can we just suppress the lines individually?

packages/default-storage-backend/package.json
```diff
@@ -84,11 +83,11 @@
     "appium-uiautomator2-driver": "^2.26.2",
     "appium-xcuitest-driver": "^4.32.5",
     "concurrently": "^6.4.0",
-    "eslint": "^8.0.0",
+    "eslint": "8.26.0",
     "eslint-plugin-wdio": "^8.8.7",
     "expo": "^48.0.0",
     "lodash": "^4.17.21",
-    "prettier": "^2.5.1",
+    "prettier": "2.4.1",
```
tido64, line 90: Why did we roll back to an older version?

packages/core/package.json
krizzu: I copied this from bob setup, so there was no file ext.
TIL about exports 😄  replaced it now

.eslintrc
krizzu: I would opt out of doing that, as this might decrease DX. IDE like Webstorm picks those config files and enables support based on their existence

packages/default-storage-backend/src/helpers.ts
krizzu: sure, I don't mind

.prettierrc
krizzu: I just checked and removing this config would replace all single quotes with doubles. If you don't mind, I can do this in this PR

tido64: Let's do it in a separate PR. I think we want to upgrade to 3.x as well.

.eslintrc
tido64: Webstorm doesn't pick the nearest config?

tido64 (approved):

.eslintrc
krizzu: You're right! I had problem with monorepo and diff eslint on Webstorm in other project, but the quick check proved it picks the nearest config!

krizzu:
@tido64 thanks for review 🙏  I'll do another PR with Prettier upgrade and moving of config files

PR #1016, opened 2023-10-19
feat(api): AsyncStorage interfaces

krizzu:
## Summary

Introducing a core API for Storage implementations - `AsyncStorage`.
This interface provides a unified and concise API, extending [Web Storage API](https://developer.mozilla.org/en-US/docs/Web/API/Storage) with additional methods for handling multiple entries at the time.

### Extension

If a Storage has capabilities beyond basic, provided by `AsyncStorage` interface, it can be delivered by extending `StorageExtension`.

### Model

The interface provides a way to define user's own model, for type safety


krizzu:
For some reason, I had YN0018 error (wrong checksum) for dependencies, so had to re-run yarn install fresh.

Additionally, what has been done in this PR:
- `core` is now named `api`
- `default-storage-backend` renamed to `default-storage` (we got rid of storage backend term)

krizzu:
For review: main change is within `api` package, everything else is adjustment for package name changes.
@tido64 This one is good for review/discussion

packages/api/example/ExampleStorage.ts
```diff
@@ -0,0 +1,71 @@
...
+  };
+
+  getMany = async <K extends StorageKeys<MyModel>>(
+    keys: K[]
+  ): Promise<{ [k in K]: MyModel[k] }> => {
+    return keys.reduce((entries, key) => {
+      return {
+        ...entries,
+        [key]: this.storage[key] ?? null,
+      };
```
tido64, line 43: Nit: We should mutate instead of copying in a loop.

packages/api/src/AsyncStorage.ts
```diff
@@ -0,0 +1,74 @@
+import { StorageKeys, StorageModel } from "./StorageModel";
+import { StorageExtension } from "./StorageExtension";
```
tido64, line 2:
```suggestion
import type { StorageKeys, StorageModel } from "./StorageModel";
import type { StorageExtension } from "./StorageExtension";
```

packages/api/src/index.ts
```diff
@@ -0,0 +1,3 @@
+export { AsyncStorage } from "./AsyncStorage";
+export { StorageExtension } from "./StorageExtension";
+export * from "./StorageModel";
```
tido64, line 3: Prefer explicit exports as it will aid tree shaking, and in this particular case, help bundlers understand that these are only types:
```suggestion
export type { StorageExtension } from "./StorageExtension";
export type { StorageKeys, StorageModel } from "./StorageModel";
```

yarn.lock
```diff
@@ -515,7 +524,53 @@ __metadata:
   languageName: node
   linkType: hard
 
-"@babel/core@npm:^7.13.16, @babel/core@npm:^7.14.0, @babel/core@npm:^7.18.5, @babel/core@npm:^7.18.6, @babel/core@npm:^7.19.6, @babel/core@npm:^7.20.0":
+"@babel/core@npm:^7.11.6, @babel/core@npm:^7.12.3":
```
tido64, line 527: Can you run `npx yarn-deduplicate` to get rid of the dupes?

tido64 (approved):
This makes sense to me. Thanks for working on it. I left a few nits only.

packages/api/src/index.ts
krizzu: make total sense 🙏

packages/api/example/ExampleStorage.ts
tido64: Instead of using the spread op every iteration:
```suggestion
    return keys.reduce((entries, key) => {
      entries[key] = this.storage[key] ?? null;
      return entries;
```

PR #1021, opened 2023-10-31
[tvOS] Apple TV warning should only happen once

douglowder:
## Summary

It is not necessary for Apple TV to do a logbox warning about async storage every time the app is reloaded.

This change just places the warning in a `dispatch_once()` so that it occurs only when the app is started.

## Test Plan

Tested using the app at https://github.com/douglowder/IgniteTVTest, which has the patch provided in this PR.


tido64 (approved):
Thanks, this looks good to me. Before we merge though, I just want to verify with @krizzu that we can still publish new versions in the current repo state.

krizzu:
Should be good to go

douglowder:
Thanks very much @tido64 @krizzu !

Issue #982, opened 2023-07-03
CocoaPods could not find compatible versions for pod "RNCAsyncStorage"


ravindraguptacapgemini:
### What happened?

After the recent update of 1.19.0, the pod install command started to fail.

Analyzing dependencies
Fetching podspec for `DoubleConversion` from `../node_modules/react-native/third-party-podspecs/DoubleConversion.podspec`
Fetching podspec for `RCT-Folly` from `../node_modules/react-native/third-party-podspecs/RCT-Folly.podspec`
RNFBApp: Using overridden static_framework value of 'true'
RNFBMessaging: Using overridden static_framework value of 'true'
Fetching podspec for `boost` from `../node_modules/react-native/third-party-podspecs/boost.podspec`
Fetching podspec for `glog` from `../node_modules/react-native/third-party-podspecs/glog.podspec`
[!] CocoaPods could not find compatible versions for pod "RNCAsyncStorage":
  In Podfile:
    RNCAsyncStorage (from `../node_modules/@react-native-async-storage/async-storage`)

Specs satisfying the `RNCAsyncStorage (from `../node_modules/@react-native-async-storage/async-storage`)` dependency were found, but they required a higher minimum deployment target.

### Version

1.19.0

### What platforms are you seeing this issue on?

- [ ] Android
- [X] iOS
- [ ] macOS
- [ ] Windows
- [ ] web

### System Information

```true
RCT_NEW_ARCH_ENABLED=1 pod install
```

### Steps to Reproduce

Just run pod install command

tido64:
What is `IPHONEOS_DEPLOYMENT_TARGET` currently set to? If you enabled New Arch, you will also need to bump it to at least 13.4:

https://github.com/react-native-async-storage/async-storage/blob/d30a69f2eb2c77a8e41a82129a1e5729c1efb015/RNCAsyncStorage.podspec#L26

Sowed:
Thank you so much @tido64

felipegmg:
Is there any problem if I need to downgrade the iOS platform to 13.0?

jorgegvallejo:
> What is `IPHONEOS_DEPLOYMENT_TARGET` currently set to? If you enabled New Arch, you will also need to bump it to at least 13.4:
> 
> https://github.com/react-native-async-storage/async-storage/blob/d30a69f2eb2c77a8e41a82129a1e5729c1efb015/RNCAsyncStorage.podspec#L26

[...]

PR #1024, opened 2023-11-06
Set iOS target to 12.4 for new architecture

hsjoberg:
## Summary

Hey, I assume this was a typo as react-native itself sets 12.4 as the minimum required version.
Without this change, you'll get a build error in the default react-native template and configuration.


krizzu:
yeah, I think it was a typo - @tido64 can you please confirm ?

tido64:
I can't recall why I asked for 13.4, but I'm pretty sure it was intentional since I've been telling people to bump their minimum deployment version: https://github.com/react-native-async-storage/async-storage/issues/982#issuecomment-1618938674

It will be default starting with 0.73 though, so we should leave this be: https://github.com/facebook/react-native/commit/610b14e4f3f606cd4c49518b9a42e3290fd52aeb

Are you having specific issues?

hsjoberg:
I see, I was not aware that react- native 0.73 is going to bump up target to iOS 13.4. Then this change does not make sense moving forward. Closing this one.

> Are you having specific issues?

Well, it's easily fixable of course, but the default react-native template \w new arch enabled will fail (<=0.72).

PR #1028, opened 2023-11-18
chore(android): Bump version for Next storage

krizzu:
## Summary

- Bumped default Kotlin version to 1.9.20
- Coroutines bumped to 1.7.3
- Dropped `kapt` in favor of `ksp`, which version can be set via gradle property

### DIFF
--- a/packages/default-storage/android/build.gradle
+++ b/packages/default-storage/android/build.gradle
@@ -19,12 +19,10 @@ buildscript {
 
 
 apply plugin: 'com.android.library'
-apply from: "config.gradle"
+apply from: 'config.gradle'
 
 boolean isNewArchitectureEnabled = ext.AsyncStorageConfig.isNewArchitectureEnabled
 boolean useNextStorage = ext.AsyncStorageConfig.useNextStorage
-
-println("[AsyncStorage] Config: ${ext.AsyncStorageConfig}")
 
 if (useNextStorage) {
     apply plugin: 'com.google.devtools.ksp'
@@ -103,26 +101,20 @@ repositories {
 dependencies {
     if (useNextStorage) {
         def room_version = project.ext.AsyncStorageConfig.roomVersion
-        def coroutines_version = "1.7.3"
-        def junit_version = "4.13.2"
-        def robolectric_version = "4.11.1"
-        def truth_version = "1.1.3"
-        def androidxtest_version = "1.4.0"
-        def androidtest_junit_version = "1.1.3"
 
         implementation "androidx.room:room-runtime:$room_version"
         implementation "androidx.room:room-ktx:$room_version"
         ksp "androidx.room:room-compiler:$room_version"
 
-        implementation "org.jetbrains.kotlinx:kotlinx-coroutines-android:$coroutines_version"
+        implementation project.ext.AsyncStorageLibs.coroutines
 
-        testImplementation "junit:junit:$junit_version"
-        testImplementation "androidx.test:runner:$androidxtest_version"
-        testImplementation "androidx.test:rules:$androidxtest_version"
-        testImplementation "androidx.test.ext:junit:$androidtest_junit_version"
-        testImplementation "org.robolectric:robolectric:$robolectric_version"
-        testImplementation "com.google.truth:truth:$truth_version"
-        testImplementation "org.jetbrains.kotlinx:kotlinx-coroutines-test:$coroutines_version"
+        testImplementation project.ext.AsyncStorageLibs.testCoroutines
+        testImplementation project.ext.AsyncStorageLibs.testJunit
+        testImplementation project.ext.AsyncStorageLibs.testExtJunit
+        testImplementation project.ext.AsyncStorageLibs.testRunner
+        testImplementation project.ext.AsyncStorageLibs.testRules
+        testImplementation project.ext.AsyncStorageLibs.testRobolectric
+        testImplementation project.ext.AsyncStorageLibs.testTruth
     }
 
     implementation 'com.facebook.react:react-native:+' // from node_modules
--- a/packages/default-storage/android/build.gradle
+++ b/packages/default-storage/android/build.gradle
@@ -105,7 +105,7 @@ dependencies {
         def room_version = project.ext.AsyncStorageConfig.roomVersion
         def coroutines_version = "1.7.3"
         def junit_version = "4.13.2"
-        def robolectric_version = "4.7.3"
+        def robolectric_version = "4.11.1"
         def truth_version = "1.1.3"
         def androidxtest_version = "1.4.0"
         def androidtest_junit_version = "1.1.3"
--- a/packages/default-storage/android/build.gradle
+++ b/packages/default-storage/android/build.gradle
@@ -1,92 +1,38 @@
-import java.nio.file.Paths
-
-def resolveModulePath(packageName) {
-    def basePath = rootDir.toPath().normalize()
-
-    // Node's module resolution algorithm searches up to the root directory,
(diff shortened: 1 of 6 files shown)


packages/default-storage-backend/android/build.gradle
```diff
@@ -47,14 +47,19 @@ buildscript {
             ? rootProject.ext['kotlinVersion']
             : rootProject.hasProperty('AsyncStorage_kotlinVersion')
             ? rootProject.properties['AsyncStorage_kotlinVersion']
-            : '1.8.10'
+            : '1.9.20'
```
tido64, line 50: Will this cause issues when `react-native` is still on 1.8? And what about older versions?

In `react-native-test-app`, we keep a list of compatible versions: https://github.com/microsoft/react-native-test-app/blob/e43b6a622bffaa76ce28e1ed1c4f036db6611342/android/dependencies.gradle#L6

Should we be doing the same here?

krizzu: Kotlin is being kept forward binary compatible, so It should be fine with older versions. Gradle will select latest kotlin version and evict the rest.

I do like your suggestion tho, I could add a map of supported KSP versions, so we can decide on it based on Kt version used

LFMAKER: The addition of the supported KSP versions map will be particularly valuable for legacy projects, providing a clear understanding of compatibility.

krizzu:
@tido64 CI is failing due to
> Plugin [id: 'com.google.devtools.ksp', version: 'null'] was not found in any of the following sources
I had same issue locally, where `dependencies.gradle` file could not be read from TestApp.

tido64:
> @tido64 CI is failing due to
> 
> > Plugin [id: 'com.google.devtools.ksp', version: 'null'] was not found in any of the following sources
> 
> I had same issue locally, where `dependencies.gradle` file could not be read from TestApp.

Seems like we're missing the latest version: https://github.com/microsoft/react-native-test-app/pull/1701

I've also updated the logic to not return `null`.

krizzu:
Bumped, let's give it a go 🤞

tido64:
Looks like we need to bump Robolectric:

```
com.reactnativecommunity.asyncstorage.next.AsyncStorageAccessTest > initializationError FAILED
    java.lang.IllegalArgumentException: failed to configure com.reactnativecommunity.asyncstorage.next.AsyncStorageAccessTest.performsBasicGetSetRemoveOperations: Package targetSdkVersion=33 > maxSdkVersion=31
        at org.robolectric.RobolectricTestRunner.getChildren(RobolectricTestRunner.java:260)
```

Latest version should support up to 34.

krizzu:
Yeah, will try to do it later day or tomorrow 🙏

packages/default-storage/android/build.gradle
```diff
@@ -1,87 +1,38 @@
...
-def useDedicatedExecutor = getFlagOrDefault('AsyncStorage_dedicatedExecutor', false)
+apply plugin: 'com.android.library'
+apply from: "config.gradle"
 
-// Use next storage implementation
-def useNextStorage = getFlagOrDefault("AsyncStorage_useNextStorage", false)
+boolean isNewArchitectureEnabled = ext.AsyncStorageConfig.isNewArchitectureEnabled
+boolean useNextStorage = ext.AsyncStorageConfig.useNextStorage
+
+println("[AsyncStorage] Config: ${ext.AsyncStorageConfig}")
```
tido64, line 27: Do we intend to keep this?

packages/default-storage/android/build.gradle
```diff
@@ -143,43 +94,36 @@ android {
...
-        // used kotlin
-        def kotlinReflect_version = project.ext.asyncStorageKtVersion
+        def room_version = project.ext.AsyncStorageConfig.roomVersion
+        def coroutines_version = "1.7.3"
         def junit_version = "4.13.2"
-        def robolectric_version = "4.7.3"
+        def robolectric_version = "4.11.1"
         def truth_version = "1.1.3"
         def androidxtest_version = "1.4.0"
         def androidtest_junit_version = "1.1.3"
```
tido64, line 111: I think we should move these into `config.gradle` as well. That way we've got all version numbers in a single place. Alternatively, if we set it up similarly to RNTA's `dependencies.gradle`, we can get Dependabot to automatically bump the version numbers as well.

packages/default-storage/android/config.gradle
```diff
@@ -0,0 +1,106 @@
...
+            "1.7.20-1.0.8",
+            "1.7.10-1.0.6",
+            "1.7.0-1.0.6",
+            "1.6.21-1.0.6",
+            "1.6.20-1.0.5",
+            "1.6.10-1.0.4",
+            "1.6.0-1.0.2",
+            "1.5.31-1.0.1",
+            "1.5.30-1.0.0",
+    ].find { it.startsWith(kotlinVersion) }
```
tido64, line 63: We should have a fallback in case a newer version of Kotlin is used:
```suggestion
    def availableVersions = [
            "1.9.20-1.0.14",
            "1.9.10-1.0.13",
            "1.9.0-1.0.13",
            "1.8.22-1.0.11",
            "1.8.21-1.0.11",
            "1.8.20-1.0.11",
            "1.8.10-1.0.9",
            "1.8.0-1.0.9",
            "1.7.22-1.0.8",
            "1.7.21-1.0.8",
            "1.7.20-1.0.8",
            "1.7.10-1.0.6",
            "1.7.0-1.0.6",
            "1.6.21-1.0.6",
            "1.6.20-1.0.5",
            "1.6.10-1.0.4",
            "1.6.0-1.0.2",
            "1.5.31-1.0.1",
            "1.5.30-1.0.0",
    ]
    def version = availableVersions.find { it.startsWith(kotlinVersion) }
    return version ?: availableVersions[0]
```

krizzu: I like fallback approach 👍

packages/default-storage/android/build.gradle
krizzu: Nice, moved it to `config.gradle`

packages/default-storage/android/build.gradle
krizzu: Removed it, but I think it might come useful for debugging purposes if we keep it. WDYT?

tido64: That makes sense. Can we use [`logger.info`](https://docs.gradle.org/current/userguide/logging.html#sec:sending_your_own_log_messages) to log this info?

tido64 (approved):
🚀

packages/default-storage/package.json
```diff
@@ -8,9 +8,9 @@
   "types": "lib/typescript/index.d.ts",
   "files": [
     "RNCAsyncStorage.podspec",
-    "android/build.gradle",
-    "android/src",
-    "android/testresults.gradle",
+    "android/",
+    "!android/.gradle",
+    "!android/build",
```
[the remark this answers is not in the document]
tido64: This might be related to the rewrite of the glob implementation in npm 9. We've hit issues with it as well: https://github.com/npm/npm-packlist/issues/152

packages/default-storage/android/build.gradle
krizzu: 👍

packages/default-storage/package.json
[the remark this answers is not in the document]
krizzu: not cool!

krizzu:
@tido64 added a log to it. I'm going to release this as minor version update, due to libraries bump

PR #1033, opened 2023-11-22
chore: bump ESLint and migrate to new config format

tido64:
## Summary

Bump ESLint and migrate to new config format

## Test Plan

CI should pass.


package.json
```diff
@@ -6,10 +6,10 @@
   ],
   "packageManager": "yarn@3.4.1",
   "devDependencies": {
-    "@react-native-community/eslint-config": "3.2.0",
-    "eslint": "8.26.0",
-    "prettier": "2.8.8",
-    "typescript": "4.9.5"
+    "@rnx-kit/eslint-plugin": "^0.5.0",
```
krizzu, line 9: why not continue with `@react-native/eslint-config` ?

package.json
```diff
@@ -6,10 +6,10 @@
...
+    "eslint": "^8.26.0",
+    "prettier": "^2.8.8",
+    "typescript": "^5.0.0"
```
krizzu, line 12: can we jump to 5.3?

packages/eslint-config/package.json
```diff
@@ -0,0 +1,27 @@
+{
```
krizzu, line 1: Couldn't we just set it in root `eslintrc`, then extend each workspace's eslint with it, instead of having a new workspace?

package.json
tido64: My issue with it is that it's doesn't use a preset. It is completely custom: https://github.com/facebook/react-native/blob/main/packages/eslint-config-react-native/index.js

Here's an example: They recently enabled a rule: https://github.com/facebook/react-native/pull/41272

This rule is enabled by default if you use the recommended preset: https://eslint.org/docs/latest/rules/no-global-assign

`@rnx-kit/eslint-plugin` is basically extending four recommended presets plus some rules for better tree shaking: https://github.com/microsoft/rnx-kit/blob/main/packages/eslint-plugin/src/configs/recommended.js

package.json
tido64: This is resolves to 5.3 in `yarn.lock`. Or did you want to explicitly specify 5.3 here? We can do that as well if you want to.

packages/eslint-config/package.json
tido64: I think the new config format recommends that you have a `eslint.config.js` in each workspace to ensure that plugins are correctly resolved (especially in a pnpm setup). This internal package will ensure that we can share the same config as we add more packages.

package.json
krizzu: thanks for in-depth take on it 🙏 make more sense now

package.json
krizzu: Yes please, I prefer to see "base" version set in package.json, rather than checking it in yarn.lock

krizzu (approved):
Thanks for that 🙏

packages/eslint-config/recommended-wdio.js
```diff
@@ -1,18 +1,16 @@
...
 module.exports = [
-  ...compat.extends("plugin:@rnx-kit/recommended"),
   {
     plugins: {
-      wdio,
+      wdio: require("eslint-plugin-wdio"),
     },
   },
-  ...compat.config(wdio.configs.recommended),
+  ...compat.extends("plugin:@rnx-kit/recommended", "plugin:wdio/recommended"),
```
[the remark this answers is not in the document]
tido64: This is a separate config file:

```js
// packages/api/eslint.config.js
module.exports = require("@react-native-async-storage/eslint-config");

// packages/default-storage/eslint.config.js
module.exports = require("@react-native-async-storage/eslint-config/recommended-wdio");
```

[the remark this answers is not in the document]
krizzu: Sounds good

krizzu (approved):

PR #1035, opened 2023-11-28
chore: fix `tsc` not checking `src/*`

tido64:
## Summary

`main` currently fails because of TypeScript errors not picked up during PR.

## Test Plan

CI should pass.


krizzu (approved):
We could move prettier to the root, but otherwise looks good 👍

PR #1064, opened 2024-02-19
fix(ios): Merging empty object should not override current value

### LINKED ISSUE #1046: `multiMerge` will *overwrite* key on iOS for '{}'
### What happened?

According to documentation, the function of `multiMerge` is `Multiple merging of existing and new values in a batch. Assumes that values are stringified JSON. Once completed, invokes callback with errors (if any).`

One expects that the empty object `{}` is the identity of the operation of merging objects, and that merging the JSON-ified object `"{}"` into any value that is already an object will produce no effect.

This is the case on Android and *not* the case on iOS (tested with 1.19.3). The following snippet reproduces this, producing `'{ "x": 1 }'` for key "a" on Android, and producing `'{}'` on iOS.

```
let { default: { multiMerge, getAllKeys, multiGet } } = require ('@react-native-async-storage/async-storage')

Promise .resolve ()
.then (_ => multiMerge ([ [ 'a', '{ "x": 1 }' ] ] ))
.then (_ => multiMerge ([ [ 'a', '{}' ] ] ))
.then (_ => getAllKeys ())
.then (keys => multiGet (keys))
.then (entries => Object .fromEntries (entries))
.then (result => console .log (result))
```

The bug is caused by this line
https://github.com/react-native-async-storage/async-storage/blob/9d9bb1aa157ce77c362e6f24639c9f2995f56e15/packages/default-storage-backend/ios/RNCAsyncStorage.mm#L791 Presumably, the line is intended to overwrite the previous value if the provided input is not an object (i.e. not mergeable). What is actually specified is overwriting of the previous value is the provided input has no keys, which unfortunately produces this bug. The corresponding implementation in Android https://github.com/react-native-async-storage/async-storage/blob/9d9bb1aa157ce77c362e6f24639c9f2995f56e15/packages/default-storage-backend/android/src/main/java/com/reactnativecommunity/asyncstorage/next/StorageSupplier.kt#L61 properly (it seems) merges the values instead, and handles the case correctly.

### Version

1.19.3

### What platforms are you seeing this issue on?

- [ ] Android
- [X] iOS
- [ ] macOS
- [ ] Windows
- [ ] web

### System Information

```true
.
```

### Steps to Reproduce

Provided above

krizzu:
## Summary

Passing an empty object to `multiMerge` avoided looping over keys to compare/merge. If the provided object is empty now, simply use the original object instead.

Closes #1046

### DIFF
--- a/packages/default-storage/README.md
+++ b/packages/default-storage/README.md
@@ -35,7 +35,7 @@ to learn more.
 2. Build app and run tests
    ```shell
    yarn bundle:ios
-   (cd example/ios && pod install)
+   pod install --project-directory=example/ios
    yarn build:e2e:ios
    yarn test:e2e:ios
    ```
--- a/packages/default-storage/README.md
+++ b/packages/default-storage/README.md
@@ -35,6 +35,7 @@ to learn more.
 2. Build app and run tests
    ```shell
    yarn bundle:ios
+   (cd example/ios && pod install)
    yarn build:e2e:ios
    yarn test:e2e:ios
    ```
--- a/packages/default-storage/example/examples/tests.ts
+++ b/packages/default-storage/example/examples/tests.ts
@@ -50,7 +50,7 @@ const tests: Record<string, TestStep[]> = {
       },
     },
   ],
-  "Should merge items #2": [
+  "Should keep existing entries when merging with an empty object": [
     {
       command: "set",
       key: "merge_test_2",
--- a/packages/default-storage/example/examples/tests.ts
+++ b/packages/default-storage/example/examples/tests.ts
@@ -13,7 +13,7 @@ export type TestStep =
     };
 
 const tests: Record<string, TestStep[]> = {
-  "Should store value in AsyncStorage": [
+  "Should store value": [
     { command: "set", key: "a", value: "0" },
     { command: "set", key: "a", value: "10" },
     { command: "set", key: "a", value: "20" },
@@ -31,8 +31,7 @@ const tests: Record<string, TestStep[]> = {
       value: {
         name: "Jerry",
         age: "21",
-        eyesColor: "blue",
-        shoeSize: "9",
+        shoeSize: "10",
       },
     },
     {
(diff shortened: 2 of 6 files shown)


packages/default-storage/ios/RNCAsyncStorage.mm
```diff
@@ -787,7 +787,8 @@ - (BOOL)_passthroughDelegate
             if (value) {
                 NSError *jsonError;
                 NSMutableDictionary *mergedVal = RCTJSONParseMutable(value, &jsonError);
-                if (RCTMergeRecursive(mergedVal, RCTJSONParse(entry[1], &jsonError))) {
+                NSDictionary *mergingValue = RCTJSONParse(entry[1], &jsonError);
+                if (!mergingValue.count || RCTMergeRecursive(mergedVal, mergingValue)) {
```
tido64, line 791: Can you also add a test case in [`packages/default-storage/example/examples/tests.ts`](https://github.com/react-native-async-storage/async-storage/blob/main/packages/default-storage/example/examples/tests.ts)?

krizzu: yeah, turns out merging tests were giving false positives, fixed that up in next commit

krizzu:
Both platform passing

Screenshot 2024-02-20 at 11 52 34

packages/default-storage/example/ios/AsyncStorageExample/AsyncStorageDevSupport.m
krizzu, line 1: Had to fix the delegate as well to actually merge dictionaries

packages/default-storage/README.md
```diff
@@ -35,6 +35,7 @@ to learn more.
 2. Build app and run tests
    ```shell
    yarn bundle:ios
+   (cd example/ios && pod install)
```
tido64, line 38: Nit:
```suggestion
   cd example/ios
   pod install
```
Alternatively:
```suggestion
   pod install --project-directory=example/ios
```

packages/default-storage/example/examples/tests.ts
```diff
@@ -42,10 +41,38 @@ const tests: Record<string, TestStep[]> = {
...
+      },
+      expected: {
+        name: "Sarah",
+        age: "23",
+        eyesColor: "green",
         shoeSize: "10",
       },
     },
   ],
+  "Should merge items #2": [
```
tido64, line 53:
```suggestion
  "Should keep existing entries when merging with an empty object": [
```

tido64 (approved):
Thanks for fixing this ❤️

packages/default-storage/example/examples/tests.ts
krizzu: ah, hate that long name in test case
`Should keep existing entries when merging with an empty object with delegate` 😄 

Let's have it in tho

krizzu:
@tido64 thanks for the review!

PR #1071, opened 2024-03-14
feat(sqlite-storage): scaffold workspace

krizzu:
## Summary

Beginning to work on SQLite storage. Basic scaffold for native module + example app using RN test app.

PR #1073, opened 2024-03-14
chore: GH workflows, turbo and other chores

krizzu:
## Summary

Doing some bump/cleanup to ease some work with future storages (#1071). Things done:

- Added turbo for easier script runs (for example, `yarn test:ts` runs ts test in all workspaces with that script)
- Added root script `test:format` to run this alongside other tests in GH action matrix
- Separated GH worksflows:
  - `pull-request` workflow is "common" for all packages (runs test* scripts)
  - `default-storage` runs only when files are changed in `default-storage` workspace (native tests + release)
  - Bumped required node to LTS (20.11.1)
- Use yarn@4
- Explicitly added .prettierrc with default config, so that Webstorm can pick it up (to ease my dev env 😄)


turbo.json
tido64, line 1: Have you looked into [Nx](https://nx.dev/)? We're using it in [rnx-kit](https://github.com/microsoft/rnx-kit/blob/main/nx.json). We tried Turbo some time back and had issues with cache file permissions.

.github/actions/setup-project/action.yml
```diff
@@ -0,0 +1,19 @@
...
+    required: false
+    description: node version to install
+    default: 20.10.0 # is already cached in Ubuntu 22.04 runner
+runs:
+  using: composite
+  steps:
+    - name: Enable corepack
+      shell: bash
+      run: |
+        corepack enable
```
tido64, line 14: Won't this require devs to also enable corepack?

krizzu: I do enable corepack to get yarn 4 installed for GH action builds

krizzu: https://yarnpkg.com/getting-started/install

turbo.json
krizzu: I'm using turbo in other project to orchestrate builds/releases and cache works fine there. I have no strong opinion here - we don't have interwoven workspace dependencies here, so I'm fine with whatever 😄

.github/actions/setup-project/action.yml
tido64: I meant locally.

krizzu: I don't think so - I haven't had to do it, but had to do it on CI, so hard to tell

tido64: Doesn't look like it:
```
% yarn
error This project's package.json defines "packageManager": "yarn@4.1.1". However the current global version of Yarn is 1.22.21.

Presence of the "packageManager" field indicates that the project is meant to be used with Corepack, a tool included by default with all official Node.js distributions starting from 16.9 and 14.19.
Corepack must currently be enabled by running corepack enable in your terminal. For more information, check out https://yarnpkg.com/corepack.
```

krizzu: so I guess it's same case for yarn@3 then? So it seems corepack needs to be enabled. Is there any drawback to that?

tido64: Unless it's enabled by default, I don't think we should require it.

krizzu: I mean, this is for our root project setting, not workspaces - those would still use whatever the project is using. Are you concerned about contributors?

krizzu: I don't think it's something you want to turn off - from node page:
>  will identify whatever package manager is configured for the current project, download it if needed, and finally run it.
So it's a package manager for package managers really - yarn 4 being part of it.

krizzu:
Rebased

krizzu:
@tido64 thanks man 🙏

PR #1082, opened 2024-04-08
chore(sqlite-storage): merge to `main`

krizzu:
Initially, I wanted to work on sqlite as feature branch, but now that there is file-change detection in place, I think it's no longer needed and work can be done on `main` directly.

This is indeed the work we merged in #1071


krizzu:
I cannot trigger the actions on this one for some reason

PR #1087, opened 2024-04-23
[windows] Add ReactPropertyBag option for DB path initialization

rozele:
## Summary

WinAppSDK apps do not support the current methods for setting / retrieving the DB path for the default storage in @react-native-async-storage/async-storage. Specifically, winrt::CoreApplication::Properties() causes a crashes on WinAppSDK.

To work around this, this change adds a specific namespaced key (ReactNativeAsyncStorage.Path) to the ReactPropertyBag for apps to register the overridden DB path for WinAppSDK apps (or apps that otherwise cannot use winrt::CoreApplication::Properties()).

## Test Plan

Compiled with WinAppSDK build of react-native-windows, set ReactNativeAsyncStorage.Path on the ReactPropertyBag for InstanceSettings when initializing the react-native-windows ReactHost, and confirmed the sqlite DB is written to the correct location.

### DIFF
--- a/packages/default-storage/windows/code/DBStorage.cpp
+++ b/packages/default-storage/windows/code/DBStorage.cpp
@@ -238,9 +238,13 @@ DBStorage::InitializeStorage(DBStorage::ErrorManager &errorManager) noexcept
 
     std::string path;
     try {
-        if (auto pathInspectable =
+        if (m_path != L"") {
+            path = winrt::to_string(m_path);
+#ifndef USE_WINUI3
+        } else if (auto pathInspectable =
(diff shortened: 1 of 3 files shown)


acoates-ms:
One thing I'd consider is exposing a function to set the Property rather than relying on magic strings.

ReactNativeAsyncStorage::SetDatabasePath(propertyBag, path)

Ideally that would be exposed in the idl file as a public method.  -- I dont know how strict the current API is about ensuring that all public APIs are part of the idl interface.

rozele:
Looks like something happened to the line endings, creating a huge diff. Will need to redo this commit.

PR #1088, opened 2024-04-25
fix react-native 0.74

a289459798:

### DIFF
--- a/packages/default-storage/android/src/main/java/com/reactnativecommunity/asyncstorage/AsyncStorageModule.java
+++ b/packages/default-storage/android/src/main/java/com/reactnativecommunity/asyncstorage/AsyncStorageModule.java
@@ -81,7 +81,7 @@ public final class AsyncStorageModule
   }
 
   @Override
-  public void onCatalystInstanceDestroy() {
+  public void invalidate() {
     mShuttingDown = true;
   }
 


hichemfantar:
Is there an open issue related to this?

coren-frankel:
#1106, I reckon.

packages/default-storage/android/src/main/java/com/reactnativecommunity/asyncstorage/AsyncStorageModule.java
```diff
@@ -81,7 +81,7 @@ public void initialize() {
   }
 
   @Override
-  public void onCatalystInstanceDestroy() {
+  public void invalidate() {
```
tido64, line 84: `invalidate` was introduced in 0.65: https://github.com/facebook/react-native/commit/18c8417290823e67e211bde241ae9dde27b72f17#diff-736ae7997f3da9751f0f5cf756d8ce4401190bda205c894acc77efe959bd1ecb

Our `package.json` says we still support 0.60: https://github.com/react-native-async-storage/async-storage/blob/fbd4ccf78b9a75d2231935321f7ae0c67811ff0a/packages/default-storage/package.json#L69

In order to make this change, we will also need to bump the minimum version to 0.65. This means we will have to make a major release.

And since `onCatalystInstanceDestroy` was removed in 0.74, it looks like this is the simpler path than maintaining two `AsyncStorageModule.java`.

@krizzu?

tido64:
Superseded by #1124

PR #1118, opened 2024-07-11
fix: set git credentials locally

krizzu:
## Summary

As per title, make sure it's not overridden by the changesets action


krizzu:
@tido64 I'll go ahead and merge this right away, to make sure release is out

PR #1132, opened 2024-07-22
docs: fix deploy again

krizzu:
## Summary

Website deployment fails silently, probably due to missing global git configuration.

### DIFF
--- a/packages/website/README.md
+++ b/packages/website/README.md
@@ -2,6 +2,7 @@
 
 Built using [Docusaurus 2](https://v2.docusaurus.io/).
 
+
 ## Development
 
 ### Installation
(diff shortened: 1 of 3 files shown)


packages/website/README.md
```diff
@@ -2,6 +2,7 @@
 
 Built using [Docusaurus 2](https://v2.docusaurus.io/).
 
+
```
krizzu, line 5: Added here to trigger website deployment after merge

tido64: Can we add `.github/workflows/website-deployment.yml` to paths instead to make sure it's always run if the pipeline changes?

tido64 (approved):

Issue #1137, opened 2024-08-01
An example output in API Docs seems misleading 


mighty-odewumi:
### What happened?

An example under the https://react-native-async-storage.github.io/async-storage/docs/api#multimerge site does not correctly show the use-case of the **multiGet** API used in the example. 

This could probably deter someone who's going through the docs and trying to whip up some solutions. They could worry about why they're getting a different output from what the example code gave and expend some precious time. 

### Version

1.21.0

### What platforms are you seeing this issue on?

- [ ] Android
- [ ] iOS
- [ ] macOS
- [ ] Windows
- [X] web

### System Information

```true
react-native: 0.73.6
react: 18.2.0
```

### Steps to Reproduce

- Visit https://react-native-async-storage.github.io/async-storage/docs/api#multimerge. 

- Check under the example given. The output written does not exhibit the expected behavior of the **multiGet** API.

PR #1139, opened 2024-08-21
Fixes the output example code multiMerge

mensonones:
Fixes the output example code in the multiMerge API documentation.

## Summary

This PR fixes the incorrect code output example for multiMerge.
Link Issue: https://github.com/react-native-async-storage/async-storage/issues/1137

## Test Plan

Run sample code

### DIFF
--- a/packages/website/docs/API.md
+++ b/packages/website/docs/API.md
@@ -379,13 +379,13 @@ const multiMerge = [
 
 
 mergeMultiple = async () => {
-  let currentlyMerged
+  let parsedCurrentlyMerged
 
   try {
     await AsyncStorage.multiSet(multiSet)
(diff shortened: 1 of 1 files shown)


krizzu:
Looks good, just that this code won't run, as `parsedCurrentlyMerged` used in console.log is not reachable, due it being initialized in `try` block

mensonones:
> Looks good, just that this code won't run, as `parsedCurrentlyMerged` used in console.log is not reachable, due it being initialized in `try` block

That's true! I didn't pay attention to that.  
Thank you for the feedback!

fixed!

krizzu (approved):

krizzu:
Thanks!

PR #1145, opened 2024-09-15
feat(expo): add an Expo config plugin to simplify enabling Next Storage implementation

### LINKED ISSUE #750: Expo next storage support config plugins
### Proposal

Hello, this 6mb limit always has been a pain in the arm, so with the new EAS build system it's now possible to lift it. (According to Brent Vatne himself the creator of expo : https://expo.canny.io/feature-requests/p/ability-to-increase-async-storage-size).

It would be great to have a config plugin : https://docs.expo.dev/guides/config-plugins/
which can add support of next storage, so that 6mb limit would get lifted.

### Alternatives

The alternative of course can be not using the next storage, and the old one but be able to configure the size, that is possible if you search this text in the link I provided in the next section `Some plugins can be customized by passing an array, where the second argument is the options:`

### Implementation Details

I don't know much but I have seen other libraries do the same for example rn-firebase has done it, and many more.
But the full tutorial on that seems to be here : https://docs.expo.dev/guides/config-plugins/

### Additional Context

_No response_

hassankhan:
## Summary

Adds a config plugin to enable [Next storage implementation](https://react-native-async-storage.github.io/async-storage/docs/advanced/next/).

Fixes #750

> [!NOTE]
> As of Expo v51, this will cause issues with the KSP version lookup so users will also need to use Expo's [`BuildProperties` plugin](https://docs.expo.dev/versions/latest/sdk/build-properties/#pluginconfigtypeandroid) to set `kotlinVersion` to `1.9.23`

## Test Plan

1. First, build the plugin:

```bash
yarn turbo run build --filter @react-native-async-storage/expo-plugin
```

2. Link the newly-built package into your Expo project (`bun/npm/yarn link`)

3. Add the config plugin to your Expo project's `app.json`:

```json
{
  "expo": {
    "plugins": [
      [
        "@react-native-async-storage/expo-plugin",
        {
          "android": {
            "nextStorage": {
              "enabled": false
            }
          }
        }
      ]
    ]
  }
}
```

4. Run prebuild in your Expo project:

```bash
npx expo prebuild -p android --clean
```

5. Check `android/gradle.properties` to confirm `AsyncStorage_useNextStorage=true` is there

### DIFF
--- a/.changeset/shaggy-eggs-grin.md
+++ b/.changeset/shaggy-eggs-grin.md
@@ -1,0 +1,5 @@
+---
+"@react-native-async-storage/async-storage-expo-plugin": minor
+---
+
+Add an Expo config plugin to simplify enabling Next Storage on Android
--- a/packages/default-storage-expo-plugin/package.json
+++ b/packages/default-storage-expo-plugin/package.json
@@ -1,0 +1,53 @@
+{
+  "name": "@react-native-async-storage/async-storage-expo-plugin",
+  "version": "1.0.0",
+  "description": "Asynchronous, persistent, key-value storage system for React Native.",
+  "main": "build/index.js",
+  "types": "build/index.d.ts",
+  "files": [
+    "build/",
+    "src/",
+    "app.plugin.js"
+  ],
+  "author": "Two Doors Dev <hello@twodoors.dev>",
+  "contributors": [
+    "Hassan Khan <h@twodoors.dev> (https://github.com/hassankhan)",
+    "Bill Spooner <b@twodoors.dev> (https://github.com/sandyklark)"
+  ],
+  "homepage": "https://github.com/react-native-async-storage/async-storage#readme",
+  "license": "MIT",
+  "keywords": [
+    "react-native",
+    "react native",
+    "async storage",
+    "asyncstorage",
+    "storage",
+    "expo",
+    "expo config plugin"
+  ],
+  "repository": {
+    "type": "git",
+    "url": "https://github.com/react-native-async-storage/async-storage.git",
+    "directory": "packages/default-storage-expo-plugin"
+  },
+  "scripts": {
+    "prepack": "yarn build",
+    "prepare": "expo-module prepare",
+    "prepublishOnly": "expo-module prepublishOnly",
+    "build": "EXPO_NONINTERACTIVE=true expo-module build",
+    "clean": "expo-module clean",
+    "lint": "expo-module lint",
+    "test:ts": "yarn build --noEmit",
+    "test:lint": "yarn lint"
+  },
+  "devDependencies": {
+    "@expo/config-plugins": "^8.0.8",
+    "eslint": "^8.54.0",
+    "expo-module-scripts": "^3.5.2",
+    "prettier": "^2.8.8"
+  },
+  "peerDependencies": {
+    "expo": "^51"
+  },
+  "upstreamPackage": "@react-native-async-storage/async-storage"
+}
--- a/packages/default-storage-expo-plugin/tsconfig.json
+++ b/packages/default-storage-expo-plugin/tsconfig.json
@@ -1,0 +1,8 @@
+{
+  "extends": "expo-module-scripts/tsconfig.plugin",
+  "compilerOptions": {
+    "outDir": "build",
+    "rootDir": "src"
+  },
+  "include": ["./src"]
+}
--- a/packages/website/docs/advanced/Next.md
+++ b/packages/website/docs/advanced/Next.md
@@ -102,6 +102,51 @@ If you want to use different KSP version, you can set a property in `gradle.prop
 AsyncStorage_next_kspVersion=1.9.24-1.0.20
 ```
 
+### Configuration with Expo
+
+If you are using Expo, you can configure the feature using [config plugins](https://docs.expo.dev/guides/config-plugins/).
+
+First, install the required plugins:
+
+```bash
+expo install expo-build-properties @react-native-async-storage/async-storage-expo-plugin
+```
+
+Next, in your Expo project's `app.json`, add the plugins:
+
+```json
+{
+  "expo": {
+    "plugins": [
+      [
+        "expo-build-properties",
+        {
+          "android": {
+            "kotlinVersion": "1.9.23"
+          }
+        }
+      ],
+      [
+        "@react-native-async-storage/async-storage-expo-plugin",
+        {
+          "android": {
+            "nextStorage": {
+              "enabled": true
+            }
+          }
+        }
+      ]
+    ]
+  }
+}
+```
+
+Finally, use the Expo prebuild command to re-generate the Android project:
+
+```bash
+expo prebuild --platform android --clean
+```
+
 ### Notable changes
 
 Alongside of a warning regarding `key`/`value`, errors are thrown when:
--- a/package.json
+++ b/package.json
@@ -45,6 +45,7 @@
   "workspaces": [
     "packages/api",
     "packages/default-storage",
+    "packages/default-storage-expo-plugin",
     "packages/eslint-config",
     "packages/sqlite-storage",
     "packages/website"
--- a/packages/default-storage-expo-plugin/app.plugin.js
+++ b/packages/default-storage-expo-plugin/app.plugin.js
@@ -1,0 +1,1 @@
+module.exports = require("./build");
--- a/packages/default-storage-expo-plugin/eslint.config.js
+++ b/packages/default-storage-expo-plugin/eslint.config.js
@@ -1,0 +1,1 @@
+module.exports = require("@react-native-async-storage/eslint-config");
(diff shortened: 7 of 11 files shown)


krizzu:
This is awesome 🙏 Let's move it as a separate workspace (packages/default-storage-expo-plugin) and we'll do npm releases from there. I understand then expo users will have to install this plugin via npm to use it, correct?

krizzu:
> typically Expo plugins are included with the native module package itself
So that in `app.config.js` you just type in `@react-native-async-storage/async-storage`, because plugin is already bundled there, right?

Async storage is used outside expo as well, so I'd rather not add anything expo specific, unless necessary. Seems like an extra package to install and configured in `app.config.js` is not a big deal (especially having documentation describing process).

wdyt @tido64 ?

hassankhan:
> Async storage is used outside expo as well, so I'd rather not add anything expo specific, unless necessary.

That does make sense - my rationale for it being there is the recommendation of Expo as the framework for React Native apps ([paraphrasing the React Native devs here](https://reactnative.dev/blog/2024/06/25/use-a-framework-to-build-react-native-apps)).

Additionally, Expo config plugins can also be used for out-of-tree RN platforms; see https://github.com/byCedric/custom-prebuild-example.

tido64:
From what I've seen, the convention is to include the plugin in the same package.

However, looking at the plugin implementation, that would mean that we have to add a direct dependency on `@expo/config-plugins`. The plugin is currently importing it, but have not declared it as a dependency. This will cause issues because now we're dependent on very specific hoisting behaviours.

That said, I don't like that fact that we're now forcing all users of AsyncStorage to download `@expo/config-plugins` regardless of whether they'll use it. So I think I'm in favour of having a separate package + install.

I'm also not a fan of the option name, `useNextStorage`. This only applies to Android and it's not obvious by just looking at the name. Can we add platform prefix i.e., `android_`?

hassankhan:
Thanks for the comments @tido64, I've updated the PR accordingly.

> However, looking at the plugin implementation, that would mean that we have to add a direct dependency on @expo/config-plugins. The plugin is currently importing it, but have not declared it as a dependency. This will cause issues because now we're dependent on very specific hoisting behaviours.

My understanding is that `@expo/config-plugin` is a dependency of Expo now, so when a config plugin is installed into an Expo project, it will use the project-local version of `@expo/config-plugin`. Looking at the [config plugins here](https://github.com/expo/config-plugins) seems to confirm this.

> That said, I don't like that fact that we're now forcing all users of AsyncStorage to download @expo/config-plugins regardless of whether they'll use it. This PR adds 3000+ lines to yarn.lock.

This would only be true if `@expo/config-plugin` was a direct dependency of the plugin.

> I'm also not a fan of the option name, `useNextStorage`. This only applies to Android and it's not obvious by just looking at the name. Can we add platform prefix i.e., android_?

I used the name as described in the documentation (which already indicates that it's only for Android). Instead of prefixing it, I've nested it since there are other configurable options that I haven't added support for.

packages/default-storage-expo-plugin/package.json
```diff
@@ -0,0 +1,54 @@
+{
+  "name": "@react-native-async-storage/expo-plugin",
```
krizzu, line 2:
```suggestion
  "name": "@react-native-async-storage/async-storage-expo-plugin",
```
for future, if we ever have another expo plugin

packages/default-storage-expo-plugin/tsconfig.json
```diff
@@ -0,0 +1,9 @@
+{
+  "extends": "expo-module-scripts/tsconfig.plugin",
+  "compilerOptions": {
+    "outDir": "build",
+    "rootDir": "src"
+  },
+  "include": ["./src"],
+  "exclude": ["**/__mocks__/*", "**/__tests__/*"]
```
krizzu, line 8: nit: u don't have mocks nor tests, so u can remove this

packages/website/docs/advanced/Next.md
```diff
@@ -102,6 +102,51 @@ If you want to use different KSP version, you can set a property in `gradle.prop
...
 ```
 
+### Configuration with Expo
+
+If you are using Expo, you can configure the feature using [config plugins](https://docs.expo.dev/guides/config-plugins/).
+
+First, install the required plugins:
+
+```bash
+expo install expo-build-properties @react-native-async-storage/expo-plugin
```
krizzu, line 112:
```suggestion
expo install expo-build-properties @react-native-async-storage/async-storage-expo-plugin
```

packages/website/docs/advanced/Next.md
```diff
@@ -102,6 +102,51 @@ If you want to use different KSP version, you can set a property in `gradle.prop
...
+      [
+        "expo-build-properties",
+        {
+          "android": {
+            "kotlinVersion": "1.9.23"
+          }
+        }
+      ],
+      [
+        "@react-native-async-storage/expo-plugin",
```
krizzu, line 130:
```suggestion
        "@react-native-async-storage/async-storage-expo-plugin",
```

tido64:
> My understanding is that `@expo/config-plugin` is a dependency of Expo now, so when a config plugin is installed into an Expo project, it will use the project-local version of `@expo/config-plugin`. Looking at the [config plugins here](https://github.com/expo/config-plugins) seems to confirm this.

That part is fine. However, dependencies may not be installed in such a way that the plugin will be able to find Expo. In a standalone repo with a single app, all dependencies are installed under `node_modules` and everything is mostly fine. But in a monorepo, things get a little more complicated. For instance, dependencies could be installed like this:

```
root
├── packages
│   └── my-app
│       └── node_modules
│           ├── @react-native-async-storage/async-storage
│           └── expo
└── node_modules
```

In this specific case, the plugin will be able to find Expo. But this isn't always the case. `@react-native-async-storage/async-storage` could be hoisted like this:

```
root
├── packages
│   └── my-app
│       └── node_modules
│           └── expo
└── node_modules
    └── @react-native-async-storage/async-storage
```

In this case, the plugin will _not_ be able to find Expo. This is why we cannot rely on implicit dependencies. We _always_ declare our dependencies.

> This would only be true if `@expo/config-plugin` was a direct dependency of the plugin.

Since it looks like `expo` re-exports `@expo/config-plugins`, another option is to make `expo` an optional dependency:

```json
  "peerDependencies": {
    "expo": ">=51"
  },
  "peerDependenciesMeta": {
    "expo": {
      "optional": true
    }
  },
```

This should allow us to keep the footprint to a minimum.

packages/default-storage-expo-plugin/package.json
```diff
@@ -0,0 +1,54 @@
...
+  "repository": {
+    "type": "git",
+    "url": "https://github.com/react-native-async-storage/async-storage.git",
+    "directory": "packages/default-storage-expo-plugin"
+  },
+  "scripts": {
+    "prepack": "yarn build",
+    "prepare": "expo-module prepare",
+    "prepublishOnly": "expo-module prepublishOnly",
+    "build": "CI=true expo-module build",
```
krizzu, line 37: why set `CI` env here?

packages/default-storage-expo-plugin/.eslintrc.js
```diff
@@ -0,0 +1,2 @@
+// @generated by expo-module-scripts
+module.exports = require('expo-module-scripts/eslintrc.base.js');
```
krizzu, line 2: are there any special lint rules in this plugin? If not, let's use one we have in this repo

.changeset/shaggy-eggs-grin.md
```diff
@@ -0,0 +1,5 @@
+---
+"@react-native-async-storage/expo-plugin": minor
```
krizzu, line 2:
```suggestion
"@react-native-async-storage/async-storage-expo-plugin": minor
```

packages/default-storage-expo-plugin/package.json
hassankhan: Unfortunately `expo-module build` always starts `tsc` in watch mode, and this was to get around that. It looks like `expo-module build` will [skip watch mode when either `CI` or `EXPO_NONINTERACTIVE` is set](https://github.com/expo/expo/blob/main/packages/expo-module-scripts/bin/expo-module-build#L30C33-L32); I can switch it to `EXPO_NONINTERACTIVE` instead?

packages/default-storage-expo-plugin/tsconfig.json
hassankhan: I left this in to prevent any future mishaps if/when tests were added.

.changeset/shaggy-eggs-grin.md
krizzu, line 1: actually, I think this can be removed, as it'll be released with version specified in package.json

krizzu:
@hassankhan thanks, appreciate the changed 🙏 

I think we missed the other configuration bits from the next storage, [such as kotlin, ksp and room versions](https://react-native-async-storage.github.io/async-storage/docs/advanced/next#configuration). Can you add this in to config as well?

krizzu:
@hassankhan Hey, if you need some help with this piece, let me know

PR #1146, opened 2024-09-25
chore: bump `react-native` to 0.75

tido64:
## Summary

Bump `react-native` to 0.75

## Test Plan

CI should pass.

## Screenshots

| Android | iOS | macOS | Windows |
| :-: | :-: | :-: | :-: |
| Screenshot_1727265519 | Simulator Screenshot - iPhone 15 Pro - 2024-09-25 at 13 53 49 | Screenshot 2024-09-25 at 14 03 22 | image |

PR #1147, opened 2024-09-25
feat(windows): add New Architecture support

Yajur-Grover:
## Summary

Adds support for the New Architecture to `async-storage`. Currently, windows support for the New Architecture is experimental and may not be complete.

## Test Plan

On a Windows machine, run the following:

```
yarn
cd packages/default-storage
yarn
yarn start:windows:fabric
```
If the `yarn start:windows:fabric` command throws errors, please try the below steps (cause of error is due to packaging issue with RNTA when running `yarn react-native run-windows` on a New Architecture application - [link](https://github.com/microsoft/react-native-test-app/pull/2067#issuecomment-2165783273)):
```
yarn
cd packages/default-storage
yarn start
```
In VS 2022, open the `AsyncStorageExample.sln` file, and launch it in `Debug|x64` configuration.

## Screenshots
On Old Architecture:
image

On New Architecture:
image


packages/default-storage/windows/code/codegen/.clang-format
tido64, line 1: Do we need to check in generated code?

packages/default-storage/windows/code/ReactPackageProvider.h
```diff
@@ -20,4 +20,4 @@ namespace winrt::ReactNativeAsyncStorage::factory_implementation
     struct ReactPackageProvider
         : ReactPackageProviderT<ReactPackageProvider, implementation::ReactPackageProvider> {
     };
-}  // namespace winrt::ReactNativeAsyncStorage::factory_implementation
+}  // namespace winrt::ReactNativeAsyncStorage::factory_implementation
```
tido64, line 23: Revert

packages/default-storage/windows/packages/Microsoft.Windows.CppWinRT.2.0.200316.3/.nupkg.metadata
tido64, line 1: We probably don't want all the NuGet artifacts checked in.

tido64 (changes requested):
- There are some changes here that seem incompatible with old architecture. Have you checked that it still works?
- There are a lot of unrelated changes that need to be reverted. Especially after https://github.com/react-native-async-storage/async-storage/pull/1146 lands.

packages/default-storage/example/examples/MergeItem.tsx
```diff
@@ -175,22 +176,22 @@ function Merge(): JSX.Element {
           testID="setDelegate_button"
           title="Set native delegate"
           disabled={
-            !NativeModules["AsyncStorageTestSupport"] ||
-            !NativeModules["AsyncStorageTestSupport"].test_setDelegate
+            !TurboModuleRegistry.get('AsyncStorageTestSupport') ||
+            !TurboModuleRegistry.get('AsyncStorageTestSupport')?.test_setDelegate
```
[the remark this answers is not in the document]
Yajur-Grover: Applied.

packages/default-storage/example/examples/MergeItem.tsx
```diff
@@ -175,22 +176,22 @@ function Merge(): JSX.Element {
...
           }
         />
         <Button
           testID="unsetDelegate_button"
           title="Unset native delegate"
           disabled={
-            !NativeModules["AsyncStorageTestSupport"] ||
-            !NativeModules["AsyncStorageTestSupport"].test_unsetDelegate
+            TurboModuleRegistry.get('AsyncStorageTestSupport') ||
+            TurboModuleRegistry.get('AsyncStorageTestSupport')?.test_unsetDelegate
```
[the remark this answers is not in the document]
Yajur-Grover: Applied this change.

packages/default-storage/example/babel.config.js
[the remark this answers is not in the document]
Yajur-Grover: Reverted.

packages/default-storage/example/android/gradlew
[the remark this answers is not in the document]
Yajur-Grover: Not sure why this file is in the PR - this file is unchanged for me locally.

packages/default-storage/example/ios/AsyncStorageExample.xcworkspace/xcshareddata/IDEWorkspaceChecks.plist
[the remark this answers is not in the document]
Yajur-Grover: Do you mean I should add the file back?

packages/default-storage/example/jest.config.windows.js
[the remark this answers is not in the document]
Yajur-Grover: Removed.

packages/default-storage/example/windows/.gitignore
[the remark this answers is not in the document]
Yajur-Grover: Removed

packages/default-storage/jest.config.windows.js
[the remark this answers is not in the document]
Yajur-Grover: Removed.

packages/default-storage/metro.config.js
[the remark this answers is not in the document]
Yajur-Grover: Removed.

packages/default-storage/package.json
```diff
@@ -57,23 +58,29 @@
     "test:ts": "tsc",
     "test:e2e:android": "scripts/android_e2e.sh 'test'",
     "test:e2e:ios": "scripts/ios_e2e.sh 'test'",
-    "test:e2e:macos": "scripts/macos_e2e.sh 'test'"
+    "test:e2e:macos": "scripts/macos_e2e.sh 'test'",
+    "windows": "react-native run-windows",
+    "test:windows": "jest --config jest.config.windows.js"
```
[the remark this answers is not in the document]
Yajur-Grover: Fixed as suggested.

packages/default-storage/windows/ReactTestApp.sln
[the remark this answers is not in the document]
Yajur-Grover: Yes - part of the unused generated `ReactTestApp` directory. Removed.

packages/default-storage/windows/ExperimentalFeatures.props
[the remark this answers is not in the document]
Yajur-Grover: Deleted.

Yajur-Grover:
> * There are some changes here that seem incompatible with old architecture. Have you checked that it still works?

I tried building with the old architecture following the below steps, and was able to successfully deploy the example app (after removing the generated code from `yarn start:windows:fabric`:
```
cd packages/default-storage
yarn start:windows
```
I changed the call to `AddAttributedModules()` depending on the value of `USE_FABRIC` - I think that should maintain the same behaviour when using the old architecture.

tido64:
The 0.75 bump just merged. This PR should reduce in size significantly if you rebase now.

packages/default-storage/windows/ReactNativeAsyncStorage/ReactNativeAsyncStorage.vcxproj
```diff
@@ -113,16 +125,24 @@
...
   </ItemDefinitionGroup>
+  <ItemDefinitionGroup>
+    <ClCompile>
+      <PreprocessorDefinitions Condition="'$(UseFabric)'=='true'">USE_FABRIC;%(PreprocessorDefinitions)</PreprocessorDefinitions>
+    </ClCompile>
+  </ItemDefinitionGroup>
+  <PropertyGroup Label="UserMacros" />
   <ItemGroup>
-    <ClInclude Include="..\code\pch.h" />
+    <ClInclude Include="pch.h" />
```
tido64, line 135: Did we include another set of `pch.*`? What are we doing about the old ones?

packages/default-storage/windows/ReactNativeAsyncStorage/resource.h
tido64, line 1: Do we really need this file? It doesn't look like it provides anything.

packages/default-storage/windows/packages/Microsoft.Windows.CppWinRT.2.0.200316.3/.signature.p7s
tido64, line 1: Can you remove all files under `packages/default-storage/wikndows/packages`?

yarn.lock
tido64, line 1: Revert. There shouldn't be any changes in here.

tido64 (changes requested):

packages/default-storage/windows/ReactNativeAsyncStorage/ReactNativeAsyncStorage.vcxproj
```diff
@@ -1,101 +1,113 @@
...
   </ImportGroup>
-  <PropertyGroup Label="UserMacros" />
-  <PropertyGroup />
   <ItemDefinitionGroup>
     <ClCompile>
       <PrecompiledHeader>Use</PrecompiledHeader>
       <PrecompiledHeaderFile>pch.h</PrecompiledHeaderFile>
       <PrecompiledHeaderOutputFile>$(IntDir)pch.pch</PrecompiledHeaderOutputFile>
-      <WarningLevel>Level4</WarningLevel>
+      <WarningLevel>Level3</WarningLevel>
```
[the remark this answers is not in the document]
Yajur-Grover: I think this was automatically done when running the cpp-lib template - should we raise it back?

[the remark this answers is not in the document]
tido64: Yes, let's not make any unnecessary changes.

packages/default-storage/windows/ReactNativeAsyncStorage/ReactNativeAsyncStorage.vcxproj
Yajur-Grover: It's used by the `ReactNativeAsyncStorage.rc` file - doesn't show up in the diff, but it looks like this: 
image

packages/default-storage/windows/ReactNativeAsyncStorage/resource.h
Yajur-Grover: It's used by the `ReactNativeAsyncStorage.rc` file - doesn't show up in the diff, but it looks like this: 
image

tido64: But it's basically an empty file. Can we delete both? I don't think AsyncStorage is ever going to include any resources.

packages/default-storage/example/android/gradlew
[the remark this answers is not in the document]
Yajur-Grover: Okay - should be back to the previous state.

packages/default-storage/example/ios/AsyncStorageExample.xcworkspace/xcshareddata/IDEWorkspaceChecks.plist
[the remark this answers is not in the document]
Yajur-Grover: Reverted.

packages/default-storage/example/macos/AsyncStorageExample.xcworkspace/xcshareddata/IDEWorkspaceChecks.plist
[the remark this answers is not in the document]
Yajur-Grover: Reverted.

packages/default-storage/windows/ReactNativeAsyncStorage/ReactNativeAsyncStorage.vcxproj
Yajur-Grover: The new `pch.*` files were generated when I ran the `cpp-lib` template. The old ones are just there, I haven't included them in the project file. Should they be okay to delete?

yarn.lock
Yajur-Grover: Reverted.

packages/default-storage/windows/packages/Microsoft.Windows.CppWinRT.2.0.200316.3/.signature.p7s
Yajur-Grover: Removed.

packages/default-storage/metro.config.js
tido64, line 1: This should also be reverted. Please note that all text files must end with an EOL: https://www.baeldung.com/linux/files-end-with-newlines

packages/default-storage/src/shouldFallbackToLegacyNativeModule.ts
tido64, line 1: Revert this file. Now we have an unused import and a formatting issue.

packages/default-storage/windows/ReactNativeAsyncStorage/ReactNativeAsyncStorage.vcxproj
tido64: If they're no longer used, we should clean up.

packages/default-storage/windows/code/codegen/.clang-format
```diff
@@ -0,0 +1,2 @@
+DisableFormat: true
+SortIncludes: false
```
tido64, line 2: Add newline to the end of file: https://www.baeldung.com/linux/files-end-with-newlines

packages/default-storage/windows/ReactNativeAsyncStorage/ReactNativeAsyncStorage.vcxproj
Yajur-Grover: Removed.

packages/default-storage/windows/code/codegen/.clang-format
Yajur-Grover: This looks like a duplicate of what's in `packages/default-storage/windows/ReactNativeAsyncStorage/codegen` - are both copies necessary?

packages/default-storage/windows/ReactNativeAsyncStorage/ReactNativeAsyncStorage.vcxproj
```diff
@@ -1,101 +1,113 @@
...
+        <WindowsTargetPlatformMinVersion Condition=" '$(WindowsTargetPlatformMinVersion)' == '' ">10.0.17763.0</WindowsTargetPlatformMinVersion>
+      </PropertyGroup>
+    </Otherwise>
+  </Choose>
   <PropertyGroup Label="ReactNativeWindowsProps">
-    <ReactNativeWindowsDir Condition="'$(ReactNativeWindowsDir)' == ''">$([MSBuild]::GetDirectoryNameOfFileAbove($(SolutionDir), 'node_modules\react-native-windows\package.json'))\node_modules\react-native-windows\</ReactNativeWindowsDir>
+    <ReactNativeWindowsDir Condition="'$(ReactNativeWindowsDir)' == ''">$([MSBuild]::GetDirectoryNameOfFileAbove($(MSBuildThisFileDirectory), 'node_modules\react-native-windows\package.json'))\node_modules\react-native-windows\</ReactNativeWindowsDir>
+    <RunAutolinkCheck>false</RunAutolinkCheck>
   </PropertyGroup>
+  <Import Project="$(ReactNativeWindowsDir)\PropertySheets\External\Microsoft.ReactNative.WindowsSdk.Default.props" />
```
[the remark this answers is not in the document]
Yajur-Grover: This change was made as part of the cpp-lib template - from my understanding, using the default WindowsSDK sheet allows the minimum version to always match whatever version RNW is using.

packages/default-storage/windows/code/codegen/.clang-format
tido64: `clang-format` travels up from the current path looking for a config file. These two are in different paths so I'd say we need both:

```
packages/default-storage/windows/ReactNativeAsyncStorage/codegen/.clang-format
packages/default-storage/windows/code/codegen/.clang-format
                         ^~~ first common dir
```

packages/default-storage/windows/ReactNativeAsyncStorage/ReactNativeAsyncStorage.vcxproj
```diff
@@ -131,42 +153,27 @@
...
-    <ProjectReference Include="$(ReactNativeWindowsDir)\Microsoft.ReactNative\Microsoft.ReactNative.vcxproj">
-      <Project>{f7d32bd0-2749-483e-9a0d-1635ef7e3136}</Project>
-      <Private>false</Private>
-    </ProjectReference>
+    <Midl Include="..\code\ReactPackageProvider.idl" />
   </ItemGroup>
   <Import Project="$(VCTargetsPath)\Microsoft.Cpp.targets" />
   <ImportGroup Label="ReactNativeWindowsTargets">
-    <Import Project="$(ReactNativeWindowsDir)\PropertySheets\External\Microsoft.ReactNative.Uwp.CppLib.targets" Condition="Exists('$(ReactNativeWindowsDir)\PropertySheets\External\Microsoft.ReactNative.Uwp.CppLib.targets')" />
+    <Import Project="$(ReactNativeWindowsDir)\PropertySheets\External\Microsoft.ReactNative.Uwp.CppLib.targets" Condition="'$(UseFabric)'!='true' And Exists('$(ReactNativeWindowsDir)\PropertySheets\External\Microsoft.ReactNative.Uwp.CppLib.targets')" />
```
acoates-ms, line 165: We should work out how to use
https://github.com/microsoft/react-native-windows/pull/13945

Do we have any variables available to the vcxproj for RNW version?  -- Seems like that would be useful so that we could import Uwp.CppLib on older RNW versions, but the newer import on newer RNW versions.

If we dont have such a variable, we should add one to newer RNW versions, and we can use the lack of such a variable as indication to use the older build logic.

packages/default-storage/windows/ReactNativeAsyncStorage/ReactPackageProvider.cpp
```diff
@@ -0,0 +1,20 @@
...
+#include "ReactNativeAsyncStorage.h"
+
+using namespace winrt::Microsoft::ReactNative;
+
+namespace winrt::ReactNativeAsyncStorageAsyncStorage::implementation
+{
+
+void ReactPackageProvider::CreatePackage(IReactPackageBuilder const &packageBuilder) noexcept
+{
+  AddAttributedModules(packageBuilder, true);
```
acoates-ms, line 17: Whats the supported RNW version range of async-storage?  -- Should we put the extra "true"  args behind a version check?

packages/default-storage/windows/code/codegen/.clang-format
acoates-ms: Something seems up if we have two copies of the generated code.  hats not right.

packages/default-storage/windows/ReactNativeAsyncStorage/ReactPackageProvider.cpp
Yajur-Grover: Actually, I think this file is unused - the included file is https://github.com/react-native-async-storage/async-storage/pull/1147/files/b9f2e55486dfcbeb15f9dc6d0ab37e999913a553#diff-91c8b85a4a6cfde2508040988d613ec059855d3e4cca40423d0edc7478af9301

I'm using `USE_FABRIC` to decide whether to use the `true` arg or not.

packages/default-storage/windows/code/codegen/.clang-format
Yajur-Grover: Looking at the commit history, I think the two copies came from two different commits. The first one was from when I think I ran `codegen-windows` independently: [codegen-windows commit](https://github.com/react-native-async-storage/async-storage/pull/1147/commits/b231d749ea98785d5ec0b149ecc907fa68a7035c)

The second one was from this commit where I I ran it again: [second codegen commit](https://github.com/react-native-async-storage/async-storage/commit/beb94d56a9bd4abd3151a911c60cd109f856c49a)

packages/default-storage/example/examples/Functional.tsx
```diff
@@ -129,9 +130,9 @@ function Functional(): JSX.Element {
         }
       })
       .then(async () => {
-        const AsyncStorageTestSupport = TurboModuleRegistry.get(
-          "AsyncStorageTestSupport"
-        );
+        const AsyncStorageTestSupport = TurboModuleRegistry
+          ? TurboModuleRegistry.get("AsyncStorageTestSupport")
+          : NativeModules["AsyncStorageTestSupport"];
```
acoates-ms, line 135: TurboModuleRegistry already fallsback to NativeModules, so no need to check both.

Yajur-Grover: When calling TurboModuleRegistry directly, the `tsc` check was throwing the following error: https://github.com/react-native-async-storage/async-storage/actions/runs/11295316739/job/31491051486#step:5:26. I tried changing the call to `TurboModuleRegistry.get<TurboModule>` in case it was a typing issue, but same result.

tido64: You're getting typing issues because `TurboModule` is a generic type, it doesn't contain the two methods that are being called further down. You need to first define the types:

```typescript
type AsyncStorageDelegate = () => void;

type AsyncStorageTestSupport = {
  test_setDelegate: (delegate: AsyncStorageDelegate) => void;
  test_unsetDelegate: (delegate: AsyncStorageDelegate) => void;
};
```

Then use it like this:

```typescript
const AsyncStorageTestSupport = TurboModuleRegistry.get<AsyncStorageTestSupport>("AsyncStorageTestSupport");
```

packages/default-storage/windows/ReactNativeAsyncStorage/ReactNativeAsyncStorage.vcxproj
Yajur-Grover, line 1: @acoates-ms this [commit](https://github.com/react-native-async-storage/async-storage/pull/1147/commits/eef71c18b9742ab2516f47f56139ff047e9887d4) should make use of https://github.com/microsoft/react-native-windows/pull/13945 once that PR is backported into 0.76.

packages/default-storage/example/examples/Functional.tsx
```diff
@@ -130,9 +129,15 @@ function Functional(): JSX.Element {
         }
       })
       .then(async () => {
-        const AsyncStorageTestSupport = TurboModuleRegistry
-          ? TurboModuleRegistry.get("AsyncStorageTestSupport")
-          : NativeModules["AsyncStorageTestSupport"];
```
Yajur-Grover: @tido64 when using your example of the `AsyncStorageTestSupport` type, I was getting a weak typing issue saying `AsyncStorageTestSupport` had no properties in common with `TurboModule` - which is why I added the `getConstants` field. 

Also, I set the AsyncStorageDelegate to have a boolean parameter here so that the `AsyncStorageTestSupport.test_setDelegate(resolve);` calls in the Promises below would work. Is there a better way to do this?

package.json
```diff
@@ -40,7 +40,8 @@
     "@react-native/js-polyfills": "^0.75.0",
     "@react-native/normalize-colors": "^0.75.0",
     "@react-native/virtualized-lists": "^0.75.0",
-    "find-babel-config/json5": "^2.1.1"
+    "find-babel-config/json5": "^2.1.1",
+    "react-native-windows": "^0.75.5"
```
Yajur-Grover, line 44: @tido64 is this change a good way to make sure RNTA uses 0.75.5, instead of publishing this [PR](https://github.com/microsoft/react-native-test-app/pull/2281)?

packages/default-storage/windows/ReactNativeAsyncStorage/ReactNativeAsyncStorage.vcxproj
```diff
@@ -1,101 +1,119 @@
...
   <ImportGroup Label="PropertySheets">
     <Import Project="$(UserRootDir)\Microsoft.Cpp.$(Platform).user.props" Condition="exists('$(UserRootDir)\Microsoft.Cpp.$(Platform).user.props')" Label="LocalAppDataPlatform" />
   </ImportGroup>
-  <ImportGroup Label="PropertySheets">
+  <ImportGroup Label="PropertySheets" Condition="'$(UseFabric)'!='true'">
     <Import Project="PropertySheet.props" />
   </ImportGroup>
-  <ImportGroup Label="ReactNativeWindowsPropertySheets">
-    <Import Project="$(ReactNativeWindowsDir)\PropertySheets\External\Microsoft.ReactNative.Uwp.CppLib.props" Condition="Exists('$(ReactNativeWindowsDir)\PropertySheets\External\Microsoft.ReactNative.Uwp.CppLib.props')" />
+  <ImportGroup Label="ReactNativeWindowsPropertySheets" Condition="'$(ReactNativeWindowsCanary)'!='true' And '$([MSBuild]::VersionLessThan($(ReactNativeWindowsMinor), 76))'">
```
acoates-ms, line 95: $([MSBuild]::VersionLessThan($(ReactNativeWindowsMajor).$(ReactNativeWindowsMinor).$(ReactNativeWindowsPatch), '0.76.0'))

Otherwise if we ever get to a version 1.0, it would count as a version less than 76.

packages/default-storage/windows/ReactNativeAsyncStorage/ReactNativeAsyncStorage.vcxproj
```diff
@@ -1,101 +1,119 @@
...
+    <Import Project="$(ReactNativeWindowsDir)\PropertySheets\external\Microsoft.ReactNative.Uwp.CppLib.props" Condition="'$(UseFabric)'!='true' And Exists('$(ReactNativeWindowsDir)\PropertySheets\External\Microsoft.ReactNative.Uwp.CppLib.props')" />
+    <Import Project="$(ReactNativeWindowsDir)\PropertySheets\External\Microsoft.ReactNative.Composition.CppLib.props" Condition="'$(UseFabric)'=='true' And Exists('$(ReactNativeWindowsDir)\PropertySheets\External\Microsoft.ReactNative.Composition.CppLib.props')" />
+  </ImportGroup>
+  <ImportGroup Label="ReactNativeWindowsPropertySheets" Condition="'$(ReactNativeWindowsCanary)'=='true' Or '$([MSBuild]::VersionGreaterThanOrEquals($(ReactNativeWindowsMinor), 76))'">
```
acoates-ms, line 99: I think there is a way to do an if else, I think its called choice or something.  You should do that, so that we only include one set of the props files.

packages/default-storage/windows/ReactNativeAsyncStorage/ReactNativeAsyncStorage.vcxproj
```diff
@@ -1,101 +1,119 @@
...
-      <!--Temporarily disable cppwinrt heap enforcement to work around xaml compiler generated std::shared_ptr use -->
-      <AdditionalOptions Condition="'$(CppWinRTHeapEnforcement)'==''">/DWINRT_NO_MAKE_DETECTION %(AdditionalOptions)</AdditionalOptions>
-      <DisableSpecificWarnings>28204</DisableSpecificWarnings>
+      <DisableSpecificWarnings>4453;28204</DisableSpecificWarnings>
       <PreprocessorDefinitions>_WINRT_DLL;%(PreprocessorDefinitions)</PreprocessorDefinitions>
       <AdditionalUsingDirectories>$(WindowsSDK_WindowsMetadata);$(AdditionalUsingDirectories)</AdditionalUsingDirectories>
+      <LanguageStandard>stdcpp20</LanguageStandard>
     </ClCompile>
     <Link>
+      <AdditionalDependencies>shell32.lib;user32.lib;windowsapp.lib;%(AdditionalDependenices)</AdditionalDependencies>
```
acoates-ms, line 116: Isn't windowsapp.lib a UWP thing?  Not sure that that should be included in fabric apps.

packages/default-storage/windows/ReactNativeAsyncStorage/ReactNativeAsyncStorage.vcxproj
```diff
@@ -113,16 +131,24 @@
       <PreprocessorDefinitions>NDEBUG;%(PreprocessorDefinitions)</PreprocessorDefinitions>
     </ClCompile>
   </ItemDefinitionGroup>
+  <ItemDefinitionGroup>
+    <ClCompile>
+      <PreprocessorDefinitions Condition="'$(UseFabric)'=='true'">USE_FABRIC;%(PreprocessorDefinitions)</PreprocessorDefinitions>
```
acoates-ms, line 136: Shouldn't this be part of the props file? - Or do all apps have to do this?

Yajur-Grover: I believe this is a part of both the [cpp-app](https://github.com/microsoft/react-native-windows/blob/8a1386110bb75f7d54ec48c8ab97c289f10ba24f/vnext/templates/cpp-app/windows/MyApp/MyApp.vcxproj#L97) and the [cpp-lib](https://github.com/microsoft/react-native-windows/blob/8a1386110bb75f7d54ec48c8ab97c289f10ba24f/vnext/templates/cpp-lib/windows/MyLib/MyLib.vcxproj#L103) template, so it's done by all New Arch apps and libraries?

packages/default-storage/windows/ReactNativeAsyncStorage/ReactNativeAsyncStorage.vcxproj
Yajur-Grover: Looking at the `When` [documentation](https://learn.microsoft.com/en-us/visualstudio/msbuild/when-element-msbuild?view=vs-2022#child-elements), we cannot add `ImportGroup` as a child element of a `<When>` statement. With the current implementation, I don't think both sets of props files will be included since both conditions won't evaluate to true at the same time.

packages/default-storage/windows/ReactNativeAsyncStorage/ReactNativeAsyncStorage.vcxproj
Yajur-Grover: Also think this is from the cpp-lib template - not sure whether we should keep it or not. @jonthysell do you have any thoughts?

package.json
tido64: You don't need to add a resolution override. You can bump the package in `yarn.lock`. Since it's already been bumped, you can remove this entry only.

packages/default-storage/example/examples/Functional.tsx
tido64: You probably want this: `type AsyncStorageTestSupport = TurboModule & {`

packages/default-storage/windows/ReactNativeAsyncStorage/ReactNativeAsyncStorage.vcxproj
Yajur-Grover: I think this change has the mentioned fix? https://github.com/react-native-async-storage/async-storage/pull/1147/commits/a1d931c2fb67fe48dd3aed2d7cd454972de3a923. 

Missed a couple of the error conditions at the bottom, fixed here: https://github.com/react-native-async-storage/async-storage/pull/1147/commits/dd1dbacb86b1735702d37d439d8f5a54cc066b54.

package.json
Yajur-Grover: Done.

packages/default-storage/windows/ReactNativeAsyncStorage/ReactNativeAsyncStorage.vcxproj
jonthysell: Yes I plan on having this ready for 0.76.0, though it will probably be a new #define for customers (like RNW_NEW_ARCHITECTURE) : https://github.com/microsoft/react-native-windows/issues/13929

packages/default-storage/windows/ReactNativeAsyncStorage/ReactNativeAsyncStorage.vcxproj
jonthysell: It's still in the new arch lib template. Maybe it isn't necessary for new arch, but it might still be for old arch.

packages/default-storage/windows/ReactNativeAsyncStorage/ReactNativeAsyncStorage.vcxproj
jonthysell: that logic is in Microsoft.ReactNative.CppLib.props and Microsoft.ReactNative.CppLib.targets and will be in 0.76. At some point we just have to say, RNW fabricis soft launching in 0.76, the template in 0.76 is the supported base and 0.76 is the supported min version for fabric. Projects are welcome to use UseFabric themselves to replicate the functionality for RNW < 0.76, but I don't want that in the template.

tido64 (approved):
Code-wise, this looks good to me. I won't be able to test this on Windows this week. It would be great if you could post screenshots of both old and new arch.

Yajur-Grover:
@tido64 sounds good - have added screenshots. It looks like there is an intermittent RNTA packaging issue present here that you have previously run into: https://github.com/microsoft/react-native-test-app/pull/2067#issuecomment-2165783273. I have added instructions to build through VS 2022 which are much more consistent to run the New Architecture app.

PR #1164, opened 2024-12-01
chore: Bump dependencies + update example setup

krizzu:
## Summary

For each storage (default and sqlite), I'm "detaching" example RN app from the actual library implementation, by moving examples to be their own workspace. This makes it less noisy in terms of config files and what's actually is a part of lib or not. 

Additionally, I bumped dependencies to almost latest. Tested most of it myself, to confirm it works, but had some issues with macOS RNCore dependency - will investigate it in a while. @tido64 could I ask you to check out the Windows example, if it works? 🙏


krizzu:
@tido64 test are passing now, ready for review 🙏

packages/default-storage/react-native.config.js
tido64, line 1: Can this file be removed now?

packages/sqlite-storage/example/package.json
```diff
@@ -0,0 +1,38 @@
+{
+  "name": "react-native-example-sqlitestorage",
+  "version": "0.0.1",
+  "private": true,
+  "scripts": {
+    "android": "react-native run-android",
+    "build:android": "npm run mkdist && react-native bundle --entry-file ../../../../../../Desktop/sqlite-storage/example/index.js --platform android --dev true --bundle-output dist/main.android.jsbundle --assets-dest dist && react-native build-android --extra-params \"--no-daemon --console=plain -PreactNativeArchitectures=arm64-v8a\"",
+    "build:ios": "npm run mkdist && react-native bundle --entry-file ../../../../../../Desktop/sqlite-storage/example/index.js --platform ios --dev true --bundle-output dist/main.ios.jsbundle --assets-dest dist && react-native build-ios --scheme AwesomeLibraryExample --mode Debug --extra-params \"-sdk iphonesimulator CC=clang CPLUSPLUS=clang++ LD=clang LDPLUSPLUS=clang++ GCC_OPTIMIZATION_LEVEL=0 GCC_PRECOMPILE_PREFIX_HEADER=YES ASSETCATALOG_COMPILER_OPTIMIZATION=time DEBUG_INFORMATION_FORMAT=dwarf COMPILER_INDEX_STORE_ENABLE=NO\"",
```
tido64, line 8: Are the entry file paths correct? They look like they reference something locally on your machine?

yarn.lock
```diff
@@ -639,6 +546,29 @@ __metadata:
   languageName: node
   linkType: hard
 
+"@babel/core@npm:^7.25.0, @babel/core@npm:^7.25.2":
```
tido64, line 549: Can you run dedupe?

package.json
```diff
@@ -21,32 +21,20 @@
...
-    "@react-native/babel-plugin-codegen": "^0.75.0",
-    "@react-native/babel-preset": "^0.75.0",
-    "@react-native/codegen": "^0.75.0",
-    "@react-native/community-cli-plugin": "^0.75.0",
-    "@react-native/community-cli-plugin/@react-native-community/cli-server-api": "^14.0.0",
-    "@react-native/community-cli-plugin/@react-native-community/cli-tools": "^14.0.0",
-    "@react-native/gradle-plugin": "^0.75.0",
-    "@react-native/js-polyfills": "^0.75.0",
-    "@react-native/normalize-colors": "^0.75.0",
-    "@react-native/virtualized-lists": "^0.75.0",
```
tido64, line 42: We should still keep these but bump the version numbers. Otherwise we will get dupes.

tido64 (commented):
Should we move the example apps out of package folders entirely?

packages/sqlite-storage/example/package.json
krizzu: Yes, lots of directory moving 😄

krizzu:
@tido64 I don't mind moving examples out, but feels like they are good where they are now

packages/default-storage/example/metro.config.js
```diff
@@ -0,0 +1,18 @@
...
+ * Metro configuration
+ * https://facebook.github.io/metro/docs/configuration
+ *
+ * @type {import('metro-config').MetroConfig}
+ */
+module.exports = getConfig(getDefaultConfig(__dirname), {
+  root,
+  pkg,
+  project: __dirname,
+});
```
tido64, line 18: We should keep the contents of the previous file:
```suggestion
const { makeMetroConfig } = require("@rnx-kit/metro-config");
const path = require("node:path");

module.exports = makeMetroConfig({
  projectRoot: path.join(__dirname, "example"),
  watchFolders: [__dirname],
  resolver: {
    extraNodeModules: {
      "@react-native-async-storage/async-storage": __dirname,
    },
  },
  transformer: {
    getTransformOptions: async () => ({
      transform: {
        experimentalImportSupport: false,
        inlineRequires: false,
      },
    }),
  },
});
```
This will ensure RN works in future versions as well.

packages/default-storage/example/app.json
```diff
@@ -3,12 +3,14 @@
...
+      "appKey": "AsyncStorageExample",
+      "displayName": "AsyncStorageExample"
     }
   ],
   "resources": {
     "android": ["index.android.jsbundle"],
     "ios": ["index.ios.jsbundle"],
-    "macos": ["index.macos.jsbundle"]
+    "macos": ["index.macos.jsbundle"],
+    "windows": ["main.windows.bundle"]
```
tido64, line 14: Is this right? Why is it the only platform using `main.*`?

packages/default-storage/package.json
```diff
@@ -23,6 +23,7 @@
     "Evan Bacon <bacon@expo.io> (https://github.com/evanbacon)",
     "Tommy Nguyen <4123478+tido64@users.noreply.github.com> (https://github.com/tido64)"
   ],
+  "source": "src/index",
```
tido64, line 26: Out of curiosity, which tool is using this field?

packages/sqlite-storage/example/metro.config.js
tido64, line 1: We should copy from the other `metro.config.js`.

packages/default-storage/example/app.json
krizzu: it's because of our building scripts outputting `index.*.jsbundle`. I changed it to `main.*.bundle` to make them all align

packages/default-storage/package.json
krizzu: I found this in Bob, required to correctly find entry point to the example app

krizzu:
@tido64 thank you for throughly done reviews 🙏

packages/default-storage/example/package.json
```diff
@@ -35,7 +36,6 @@
     "@react-native-community/cli-platform-android": "^15.0.0",
     "@react-native-community/cli-platform-ios": "^15.0.0",
     "@react-native/babel-preset": "^0.76.0",
-    "@react-native/metro-config": "^0.76.0",
```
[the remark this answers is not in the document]
krizzu: just realized 😄

PR #1178, opened 2025-01-31
chore(apple): fix build issues discovered internally

Saadnajmi:
## Summary

At my side of Microsoft, we have recently "vendored" a couple of libraries. Which means, we started building them from source using our own generated Xcode project (not the one cocoapods generates), which has much stricter build flags / treats warnings as errors. In doing so, I caught a couple of issues in React Native async-storage I'd like to push back up!

Namely:
- Fix warnings where we need to add a bunch of either `__nullable` or `__nonnull` specifiers. 
- Fix a "declaration of variable shadows another local variable" where we had two `NSError *error = nil` declarations in a row
- Fix another "declaration of variable shadows another local variable" with two NSError variables by renaming one of them
- Add an `__unused` clang tag to a parameter
- Add `__autoreleaseing` clang tags to two more parameters. 
  - The error internally without these was `Method parameter of type 'NSMutableArray<NSDictionary *> *__autoreleasing *' with no explicit ownership`, for the record. 

## Test Plan

CI should pass

### DIFF
--- a/packages/default-storage/ios/RNCAsyncStorage.h
+++ b/packages/default-storage/ios/RNCAsyncStorage.h
@@ -27,6 +27,9 @@
  *
  * Keys and values must always be strings or an error is returned.
  */
+
+NS_ASSUME_NONNULL_BEGIN
+
 @interface RNCAsyncStorage : NSObject <
 #ifdef RCT_NEW_ARCH_ENABLED
                                  NativeAsyncStorageModuleSpec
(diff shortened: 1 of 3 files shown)


packages/default-storage/ios/RNCAsyncStorage.h
```diff
@@ -50,13 +50,13 @@
...
 
 // Add multiple key value pairs to the cache.
-- (void)multiSet:(NSArray<NSArray<NSString *> *> *)kvPairs
-        callback:(RCTResponseSenderBlock)callback;
+- (void)multiSet:(NSArray<NSArray<NSString *> *> *_Nullable)kvPairs
+        callback:(RCTResponseSenderBlock _Nullable )callback;
 
 // Interface for natively fetching all the keys from the storage data.
-- (void)getAllKeys:(RCTResponseSenderBlock)callback;
+- (void)getAllKeys:(RCTResponseSenderBlock _Nullable )callback;
```
tido64, line 60: I think it would be more consistent with the [JS side](https://github.com/react-native-async-storage/async-storage/blob/92cf9181c1f8769b34e6f8161b742cd155d1e2d9/packages/default-storage/src/types.ts#L44) to add `NS_ASSUME_NONNULL_BEGIN`/`NS_ASSUME_NONNULL_END` and only make the callbacks nullable.

packages/default-storage/ios/RNCAsyncStorage.h
```diff
@@ -57,6 +60,8 @@
         callback:(RCTResponseSenderBlock)callback;
 
 // Interface for natively fetching all the keys from the storage data.
-- (void)getAllKeys:(RCTResponseSenderBlock)callback;
+- (void)getAllKeys:(RCTResponseSenderBlock )callback;
```
tido64, line 63:
```suggestion
- (void)getAllKeys:(RCTResponseSenderBlock)callback;
```

tido64 (approved):

tido64:
Thanks for the fixes 🙏

tido64:
> ###  ⚠️  No Changeset found
> 
> Latest commit: abafda4f09c2d426d253565744c642fb86bad458
> 
> Merging this PR will not cause a version bump for any packages. If these changes should not result in a new version, you're good to go. **If these changes should result in a version bump, you need to add a changeset.**

Oh, can you add a change log?

PR #1279, opened 2026-03-28
fix(android): Resolve the issue where users need to manually import local_repo in build.gradle in version v3.0.2.

2Stephen:
Resolve the issue where users need to manually import local_repo in build.gradle in version v3.0.2.
So Android users don‘t need have to import `local_repo` setting about like README.

### Bug 
```bash
error Failed to install the app. Command failed with exit code 1: ./gradlew app:installDebug -PreactNativeDevServerPort=8081 FAILURE: Build failed with an exception. * What went wrong:
Could not determine the dependencies of task ':app:compileDebugJavaWithJavac'.
> Could not resolve all dependencies for configuration ':app:debugCompileClasspath'. 
> Could not find org.asyncstorage.shared_storage:storage-android:1.0.0. Required by: project ':app' 
> project :react-native-async-storage_async-storage * Try:
```
### Resolve Detail
When I imported this dependency and tried to compile it into an Android app, the system threw the error "ould not find org.asyncstorage.shared_storage:storage-android:1.0.0". I tried many solutions, but none worked. I also couldn't find the dependency org.asyncstorage.shared_storage:storage-android:1.0.0 in official repositories such as Maven Central. Finally, I checked the official repository and found that the README states **you need to declare the local_repo before building for Android** so that this dependency can be resolved correctly.

## Test Plan

`./gradlew app:assembleDebug | cat` and it can build success without import `local_repo` in users' `build.gradle`

### DIFF
--- a/packages/async-storage/android/build.gradle
+++ b/packages/async-storage/android/build.gradle
@@ -77,6 +77,16 @@ android {
     }
 }
 
+def storageRepo = "$projectDir/local_repo"
+
+rootProject.allprojects {
+    repositories {
+        maven {
+            url storageRepo
+        }
+    }
+}
+
 repositories {
     mavenCentral()
     google()


krizzu:
Closing, as v3.1.0 moved artifact to MavenCentral, meaning it is no longer required to add maven local url

PR #1299, opened 2026-05-21
feat: Distribute shared-storage via Maven Central

krizzu:
## Summary

Async Storage v3's android artifact is distributed locally, via `local_repo`. This required an extra step from end users to add maven local url.

This PR changes that, to distribute the artifact via Maven Central - removing the installation steps completely. 

## Next Step

1. Do not create changeset here
2. Open new PR, with changeset, released v1.0.0 of Shared Storage, bumped version of documentation + update installation step to remove an extra step for android


scripts/build-native-lib.sh
```diff
@@ -22,14 +21,9 @@ build_android() {
   log "👷 Assembling android shared-storage"
   ./gradlew :$MODULE_NAME:$ANDROID_BUILD_TASK
 
-  log "Publishing binaries to local repo"
+  log "Publishing binaries to mavem local"
```
krizzu: 🤦

Issue #1331, opened 2026-08-27
Reproducible batch-key, transaction and storage error-handling failures in v3.1.1


OskarEichler:
## Reproducible v3.1.1 storage failures

While reviewing the current package (`4535ff0`), I reproduced these independent issues. I am preparing focused PRs, as requested by the contribution guide.

- `getMany` drops the string key `__proto__` on web, native, legacy and Windows paths because assignment invokes the inherited object setter. The key should be an ordinary own property, without changing the result's prototype.
- A rejected IndexedDB open produces an unhandled rejection even if the caller catches it. A deletion whose request succeeds but transaction aborts resolves successfully and leaves the transaction rejection unhandled. Reproduced with the actual modules, `idb` and an isolated IndexedDB implementation.
- `AsyncStorageError.nativeError(null)` throws a TypeError, an existing AsyncStorageError is thrown instead of returned, and LegacyStorageException is misclassified as UnknownError.
- Apple's legacy storage removes a manifest entry after a file-read error and puts failed file writes into its read cache. Compiled actual Objective-C method bodies reproduce both, with synthetic read failures and temporary-directory write failures.
- Windows SQLite code truncates keys/values containing NUL, clears transaction ownership before a failed COMMIT can roll back, and asserts for empty multiGet/multiRemove batches. Actual portable C++ methods/helpers reproduce these against in-memory SQLite; a Windows build has not been performed.

No production/user storage or real credentials were accessed. Proposed fixes preserve the public signatures and successful ordinary-key behavior. Observable corrections will be documented in each PR, especially full-byte Windows strings and transaction failures now rejecting.

krizzu:
> getMany drops the string key __proto__ on web, native, legacy and Windows paths because assignment invokes the inherited object setter. The key should be an ordinary own property, without changing the result's prototype.

Can you explain in your own words what exactly the problem is here? Do I understand correctly your LLM wants to handle case where one of the key is `__proto__` ?

PR #1333, opened 2026-08-27
fix: normalize native storage errors without throwing

OskarEichler:
## Summary

Return existing AsyncStorageError instances, handle null rejection values safely, and classify LegacyStorageException as OtherStorageError.

The factory now returns (rather than throws) an existing AsyncStorageError, as its return type specifies. LegacyStorageException classification changes from Unknown to OtherStorageError.

## Test Plan

- Actual error factory: baseline throws for null and its own error, and labels LegacyStorageException Unknown; fixed returns the original error, handles null and uses OtherStorageError.


krizzu (approved):
coolio, thanks

PR #1334, opened 2026-08-27
fix(web): track transaction completion and connection lifetime

OskarEichler:
## Summary

Propagate open/transaction failures through caller promises, wait for deletion commit, close blocking connections and discard terminated connections.

removeMany now rejects when its transaction aborts. Version-change requests close the old connection; later calls reopen (and reject normally if another client upgraded beyond version 1). Blocked opens wait through IndexedDB instead of throwing from an unrelated event callback.

## Test Plan

- Real idb 8.0.3 with fake-indexeddb: failed-open retry, get/put/delete aborts, rollback, empty batches, normal CRUD, forced termination and external database deletion. Baseline has unhandled rejections, resolves an aborted deletion, reuses a closed database and prevents external deletion; fixed handles all cases.


krizzu (commented):
this change removes thrown errors that are meaningful to user. I would rather throw an error, than silently fail with no info

OskarEichler:
Agreed that callers should retain meaningful failure information. I restored the typed `WebStorageError` from the `blocked` callback while keeping `blocking` as an active lifecycle response: the existing connection closes on `versionchange`, which is what lets the pending upgrade proceed instead of leaving it silently blocked. I also ran the repository formatter over both touched storage files. The package build and TypeScript check pass; the checked-in tree currently has no ESLint config discoverable by its ESLint 9 package script, so that local lint command cannot initialize.

krizzu:
> as an active lifecycle response: the existing connection closes on versionchange, which is what lets the pending upgrade proceed instead of leaving it silently blocked.

Yes, that's true, I believe this is also a good thing to give feedback to consumer about this happening anyway, so throwing an error in order to retry would make sense

PR #1335, opened 2026-08-27
fix(apple): retain committed legacy state after file errors

OskarEichler:
## Summary

Keep manifest entries on file read errors and cache large values only after successful writes.

A transient read error no longer forgets a stored key. Failed writes no longer expose the uncommitted replacement through cache. No data format/schema change.

## Test Plan

- Compiled actual Objective-C method bodies against Foundation: injected read failure preserves manifest; actual write failure preserves prior cached value. Successful writes/reads and genuinely missing files still follow existing behavior.


krizzu:
Thanks, but this is legacy implementation, that lives in v2 currently and I dont want to change behavior in v3

OskarEichler:
Understood—thanks for clarifying the v2/v3 boundary. I’ll leave the legacy Apple implementation unchanged.

PR #1336, opened 2026-08-27
fix(windows): preserve SQLite strings and roll back failed commits

OskarEichler:
## Summary

Use explicit SQLite string lengths, handle empty batches and retain transaction ownership until COMMIT succeeds.

Embedded-NUL strings are no longer truncated; empty reads/removals return empty/success instead of asserting; failed commits roll back. No schema change. This does not migrate values already truncated by prior writes.

## Test Plan

- Compiled actual portable C++ helpers and MultiGet/MultiSet/MultiRemove/GetAllKeys against in-memory SQLite on macOS. Embedded-NUL keys/values remain distinct and round-trip; only the requested key is removed; empty batches succeed; deferred-constraint COMMIT failure rolls back. Baseline reproduces truncation, assertion and dangling transaction.


krizzu:
again, this is legacy implementation that is currently rolling in v2. I don't want to touch it in v3

OskarEichler:
Understood—thanks for clarifying. I’ll leave the legacy Windows implementation unchanged across the v2/v3 transition.

PR #1337, opened 2026-08-27
perf(android): use indexed lookups for legacy batches

OskarEichler:
## Summary

Replace repeated linear searches and list copies in legacy reads/merges with hash lookups.

No intended API or ordering change; legacy deduplication is deliberately preserved.

## Test Plan

- Compiled actual Kotlin read bodies: 500 randomized equivalence cases preserve database order, appended missing keys, nulls and duplicate-key behavior. An illustrative 10,000-missing-key run was 167 ms before and under 1 ms after (single local run, not a production benchmark). Merge lookup uses the database's unique primary-key invariant; JSON merge behavior is unchanged.


krizzu:
Thanks again, but I dont want to change behavior of legacy storage between v2 and v3.

OskarEichler:
Understood—thanks for clarifying. I’ll leave the legacy Android batch implementation unchanged across the v2/v3 transition.

PR #1338, opened 2026-08-27
fix(android): share atomic storage creation across access paths

OskarEichler:
## Summary

Reuse getStorage's computeIfAbsent path when constructing RNStorage to prevent redundant native storage creation.

No API change; concurrent callers no longer initialize an unused database instance.

## Test Plan

- Compiled actual Kotlin registry with a controlled two-thread constructor overlap: simultaneous RN/native access creates two SharedStorage instances before, one after; both paths return the same retained instance.


krizzu (commented):
if this PR is only to "prevent redundant native storage creation.", `getOrPut` already does just that.

OskarEichler:
The specific race was across `getStorage` and `getRNStorage`: Kotlin’s `ConcurrentMap.getOrPut` may evaluate its default more than once under contention, while `computeIfAbsent` performs the creation atomically. A concurrent call through each access path can therefore construct two `SharedStorage` instances even though only one is retained. That said, I understand the decision to leave this path unchanged, so I’ll leave the PR closed.

krizzu:
yes, that'd make sense for RN if there would be concurrency concerns and need for atomicity here

PR #1339, opened 2026-08-27
fix(mock): clear held storage instances when resetting

OskarEichler:
## Summary

Clear existing mock stores before resetting the registry and preserve the default legacy instance's registry identity.

clearAllMockStorages now actually clears previously held instances. Existing named references are cleared, while a subsequent named factory call still creates a fresh instance as before. The default legacy instance stays registered.

## Test Plan

- Actual compiled shipped mock: default and named held references both retain values after baseline reset; fixed clears both and keeps default === createAsyncStorage('legacy'). Public signatures remain unchanged.


krizzu (approved):
ye, make sense

PR #1340, opened 2026-08-27
fix(build): dispatch both native targets for all

OskarEichler:
## Summary

Run Android and Apple builders for the native build script's all option.

Only the developer build script changes; no runtime API or package schema change.

## Test Plan

- sh -n and actual case-dispatcher with stub builders: android/apple unchanged; all changes from exit 127 (undefined build_all) to both builders in order; invalid targets still exit 1.


krizzu (approved):
make sense

